PR #1, opened 2019-07-10
Initial Pull Request

brentru:
Initial pull request for MiniMQTT - a featured MQTT client for CircuitPython 

This library is based off previous work performed by pfalcon on the uMQTT library (and the subsequent port to ESP32SPI by beachbc) and follows similar callback conventions and method signatures from the Eclipse Foundation's Paho.Mqtt.Python.

**MiniMQTT API Overview** 
* `connect`: initiates a connection with the mqtt broker 
* `reconnect`: handles reconnection to the mqtt broker and re-subscription to topics 
* `last_will`: Sets the client's optional last will and testament properties 
* `publish`: Publishes to a broker's mqtt topic. Handles data type conversions. 
* `subscribe`: Subscribes to a broker's mqtt topic. Can subscribe to one, or multiple topics with varying QoS levels. 
* `unsubscribe`: Unsubscribes from a broker's mqtt topic or multiple topics.
* `loop`: Non-blocking messaging loop, useful for checking incoming subscription messages.
* `loop_forever`: Blocking messaging loop

* `set_logger_level`/`log`: Attaches/modifies a logger to the MiniMQTT client


abachman:
I notice there are a lot of normal commands that can result in errors being raised, specifically related to is_connected. 

The examples should either use `try: except:` around every call or it might be worth figuring out a different way to address unexpected disconnection.

brentru:
A method of addressing this unexpected disconnection might be passing each
command method (publish, subscribe) to a high level method which checks the
connection status and handles hardware and network reconnections.

Though this does add overhead - it might be nicer than wrapping each
publish/subscribe call in a try/except loop?

abachman:
Are you strongly attached to "MiniMQTT", or would "MQTT" be okay?

Also, is it possible to make the name of the repository the same as the name of the module?

adafruit_minimqtt.py
```diff
@@ -0,0 +1,705 @@
...
+                if self._logger is not None:
+                    self._logger.debug('Reconnected to broker')
+                if resub_topics:
+                    if self._logger is not None:
+                        self._logger.debug('Attempting to resubscribe to prv. subscribed topics.')
+                    while self._subscribed_topics:
+                        feed = self._subscribed_topics.pop()
+                        self.subscribe(feed)
+            except OSError as e:
+                print('Failed to connect to the broker, retrying\n', e)
```
ladyada: should be logged not printed?

adafruit_minimqtt.py
```diff
@@ -0,0 +1,705 @@
...
+        """Pings the MQTT Broker to confirm if the broker is alive or if
+        there is an active network connection.
+        """
+        self.is_connected()
+        if self._logger is not None:
+            self._logger.debug('Sending PINGREQ')
+        self._sock.write(MQTT_PINGREQ)
+        if self._logger is not None:
+            self._logger.debug('Checking PINGRESP')
+        while 1:
```
ladyada: while true: is more pythony

adafruit_minimqtt.py
```diff
@@ -0,0 +1,705 @@
...
+            op = self._wait_for_msg(0.5)
+            if op == const(208):
```
ladyada: you can just have `208` here

adafruit_minimqtt.py
```diff
@@ -0,0 +1,705 @@
...
+                ping_resp = self._sock.read(2)
+                if ping_resp[0] != const(0x00):
```
ladyada: ditto, `0x00`

brentru:
@abachman  MQTT would not be OK as ReadTheDocs requires a non-generic slug url (i.e: https://readthedocs.org/projects/mqtt/ is already taken).

ladyada:
i think this is good for an initial version, and we'll see how it goes!

tannewt:
Shouldn't the readthedocs slug be `adafruit-circuitpython-mqtt` (or `adafruit-circuitpython-minimqtt`)? It's only as a circuitpython subproject that it's shortened to `mqtt` or `minimqtt`.

brentru:
@tannewt Yep, just redid the RTD setup to use that slug. RTD badge on README is now green.

tannewt:
Perfect! Glad it helped.

PR #12, opened 2019-10-16
Expose WiFi Attribute

brentru:
Exposing wifi attribute so wrapper libraries (AWS IoT, Adafruit IO) which create a `minimqtt` object can access WiFiManager and ESP32SPI (via `client.wifi.esp`) attributes.

### DIFF
--- a/adafruit_minimqtt.py
+++ b/adafruit_minimqtt.py
@@ -100,12 +100,12 @@ class MQTT:
         self._socket = socket
         network_manager_type = str(type(network_manager))
         if 'ESPSPI_WiFiManager' in network_manager_type:
-            self._wifi = network_manager
+            self.wifi = network_manager
         else:
             raise TypeError("This library requires a NetworkManager object.")
         # broker
         try: # set broker IP
-            self.broker = self._wifi.esp.unpretty_ip(broker)
+            self.broker = self.wifi.esp.unpretty_ip(broker)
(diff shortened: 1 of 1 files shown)


brentru:
@ladyada  this only effects the check for a `ESPSPI_WiFiManager`. Other `network_manager`'s (when we make them) would be detected and handled separately. 


```
        network_manager_type = str(type(network_manager))
        if 'ESPSPI_WiFiManager' in network_manager_type:
            self.wifi = network_manager
        else:
            raise TypeError("This library requires a NetworkManager object.")
```

ladyada:
yeah but if its exposed, then people would rely on it?

brentru:
@ladyada  I don't think user-code will rely on this....Possibly a better method would be checking the `esp`'s version from code if an `esp` is used, instead of relying on the MiniMQTT client.

brentru:
@ladyada I don't think exposing from user-code would be a good idea, since libraries using a MiniMQTT object should be able to check `set_psk` and `set_crt` booleans (https://github.com/adafruit/Adafruit_CircuitPython_AWS_IOT/blob/master/adafruit_aws_iot.py#L82) on initialization...

I _could_ pass in the `wifi` object to the library's init instead.

brentru:
Closing in favor of https://github.com/adafruit/Adafruit_CircuitPython_AWS_IOT/pull/4

Issue #16, opened 2019-12-28
.encode not available on STM32F05 build -- potential workaround


jerryneedell:
The STM32F405 build does not have .encode available.
see this discussion https://github.com/adafruit/circuitpython/issues/2433#issuecomment-569422114

I have a fix for this -- so it works without .encode
do you want a PR for it?

here are the diffs
```
jerryneedell@Ubuntu-Macmini:~/projects/stm32f405/airlift$ diff adafruit_minimqtt.py ~/projects/adafruit_github/Adafruit_CircuitPython_MiniMQTT/
121c121
<         if self.password is not None and len(bytes(password,"utf-8")) > MQTT_TOPIC_LENGTH_LIMIT:
---
>         if self.password is not None and len(password.encode('utf-8')) > MQTT_TOPIC_LENGTH_LIMIT:
129c129
<             self.client_id = 'cpy{0}{1}'.format(microcontroller.cpu.uid[randint(0, len(microcontroller.cpu.uid)-1)],
---
>             self.client_id = 'cpy{0}{1}'.format(microcontroller.cpu.uid[randint(0, 15)],
355c355
<             msg = bytes(str(msg),"ascii")
---
>             msg = str(msg).encode('ascii')
357c357
<             msg = bytes(str(msg),"utf-8")
---
>             msg = str(msg).encode('utf-8')
690c690
<             self._sock.send(bytes(string, "utf-8"))
---
>             self._sock.send(str.encode(string, 'utf-8'))
705c705
<         elif len(bytes(topic,'utf-8')) > MQTT_TOPIC_LENGTH_LIMIT:
---
>         elif len(topic.encode('utf-8')) > MQTT_TOPIC_LENGTH_LIMIT:
```

note the change at line 129 -- the STM build microcontroller.cpu.uid is only 12 bytes so this was failing -- I set it to len() to avoid the error

jerryneedell:
regarding the length of microcontroller.cpu.uid I have found that it varies
for atmel MCUs it appear to be 16 bytes long, but for nrf52840's it is only 8! Has anyone tried this library on an nrf52840??
For the stm32f405 it is 12 bytes

so regardless of the .encode issue, it looks like line 129 should be changed.

Do you want me to create a PR for that? would you prefer 2 different PRs?
one for .encode and one for the uid fix?

jerryneedell:
@brentru The STM32F405 issue has been made moot since .encode is now available in current master.
[...]

PR #19, opened 2020-01-06
 revise generation of client ID

jerryneedell:
addresses part of  #16
use time.monotonic() instead of microcontroller.cpu.uid for random seed to create client ID

### DIFF
--- a/adafruit_minimqtt.py
+++ b/adafruit_minimqtt.py
@@ -42,7 +42,6 @@ Implementation Notes
 import struct
 import time
 from random import randint
-import microcontroller
 from micropython import const
 import adafruit_logging as logging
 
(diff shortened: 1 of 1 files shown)


jerryneedell:
Tested with grandcentral_m4 and Airlift shield

adafruit_minimqtt.py
```diff
@@ -126,8 +125,8 @@ def __init__(self, socket, broker, port=None, username=None,
             self.client_id = client_id
         else:
             # assign a unique client_id
-            self.client_id = 'cpy{0}{1}'.format(microcontroller.cpu.uid[randint(0, 15)],
-                                                randint(0, 9))
+            self.client_id = 'cpy{0}{1}'.format(randint(0, int(time.monotonic()*100)%1000),
```
brentru: @jerryneedell 
> According to MQTT V3. 1 Protocol Specification, "The Client Identifier (Client ID) is between 1 and 23 characters long

I do not think we will ever generate a CID >23 char since this step is performed during initialization in use-cases where a CircuitPython device performs a reset or at boot-up.

However, what about if a long `time.monotonic` value is returned after a delayed initialization or delayed WiFi connection (if the power in the house is down, the AP will take a while to reboot, code will retry until the modem and AP is fully online). 

Do you feel we should truncate a generated `client_id` here to 23chars, before it fails the check below and raises a `ValueError`

jerryneedell: this code will never use more than 3 digits from the time.monotonic()  so the whole thing will only be ```cpyxxxyy```  so 7 digits at most.. Do you want it longer. It was even fewer the way it was oringally done.

brentru: Nope, 1<=cid<=23, so it's valid. I'll mergein and release! Thanks for fixing this.

PR #15, opened 2019-12-19
MQTT protocol over bluetooth

Tasm-Devil:
With this commit, the MQTT class now also supports Bluetooth. I wrote a new socket class based on [Adafruit_CircuitPython_ESP32SPI](https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/c28977f75e50cbfb6b8c5b3f7747d9dd23ffdfe6/adafruit_esp32spi/adafruit_esp32spi_socket.py). There is also an example. I split the MQTT class somewhat so that there is a clear separation between MQTT and the underlying protocol.

I hope this work will be useful to others.

Warm greetings from Germany.

PR #44, opened 2020-07-18
api doc's param & code-block display correctly

2bndy5:
I went to read the docs for this library and noticed all the parameters and example code-blocks were not displaying properly because there was no empty line that preceded them. Verified locally via `sphinx-build` that the changes generate properly.

Upon further reading I noticed there was a (un-admonished) deprecation notice and a reference to an example that was not being displayed on RTD. I added the example to `examples.rst` as "Basic Forever Loop", and I changed the notice to an admonished "Note" so that it sticks out to the reader better.

Side-note: there are a lot more examples in this library that are not being displayed on RTD. Furthermore there are no module-level docstrings that would help identify the purpose of the examples (other than the example filename). I'm not too familiar with this library, so I didn't go crazy adding all the examples to `examples.rst`.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -204,8 +204,8 @@ class MQTT:
         :param int,float,str payload: Last will disconnection payload.
             payloads of type int & float are converted to a string.
         :param int qos: Quality of Service level, defaults to
-            zero. Conventional options are ``0`` (send at least once), ``1``
-            (send at most once), or ``2`` (send exactly once).
+            zero. Conventional options are ``0`` (send at most once), ``1``
+            (send at least once), or ``2`` (send exactly once).
 
             .. note:: Only options ``1`` or ``0`` are QoS levels supported by this library.
         :param bool retain: Specifies if the payload is to be retained when
@@ -418,8 +418,8 @@ class MQTT:
         :param str,int,float msg: Data to send to the broker.
         :param bool retain: Whether the message is saved by the broker.
         :param int qos: Quality of Service level for the message, defaults to
-            zero. Conventional options are ``0`` (send at least once), ``1``
-            (send at most once), or ``2`` (send exactly once).
+            zero. Conventional options are ``0`` (send at most once), ``1``
+            (send at least once), or ``2`` (send exactly once).
 
             .. note:: Only options ``1`` or ``0`` are QoS levels supported by this library.
 
@@ -524,8 +524,8 @@ class MQTT:
             element should be a tuple containing a topic identifier string and
             qos level integer.
         :param int qos: Quality of Service level for the topic, defaults to
-            zero. Conventional options are ``0`` (send at least once), ``1``
-            (send at most once), or ``2`` (send exactly once).
+            zero. Conventional options are ``0`` (send at most once), ``1``
+            (send at least once), or ``2`` (send exactly once).
 
             .. note:: Only options ``1`` or ``0`` are QoS levels supported by this library.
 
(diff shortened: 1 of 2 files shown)


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -114,7 +115,6 @@ class MQTT:
     :param bool is_ssl: Sets a secure or insecure connection with the broker.
     :param bool log: Attaches a logger to the MQTT client, defaults to logging level INFO.
     :param int keep_alive: KeepAlive interval between the broker and the MiniMQTT client.
-
```
brentru, line 117: This blank line should remain here
> Insert a blank line after all docstrings (one-line or multi-line) that document a class
 https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -193,19 +193,23 @@ def __exit__(self, exception_type, exception_value, traceback):
         self.deinit()
 
     def deinit(self):
-        """De-initializes the MQTT client and disconnects from
-        the mqtt broker.
-
+        """De-initializes the MQTT client and disconnects from the mqtt broker.
```
brentru, line 196: Can this fit on one line instead?

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -300,8 +303,7 @@ def connect(self, clean_session=True):
                 raise MMQTTException("Invalid broker address defined.", e)
 
         # Fixed Header
-        fixed_header = bytearray()
-        fixed_header.append(0x10)
+        fixed_header = bytearray([0x10])
```
brentru, line 306: Could we do this in a separate PR instead? I like it, but it shouldn't be included with this PR.

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -668,11 +692,11 @@ def loop_forever(self):
         method if you want to run a program forever.
         Code below a call to this method will NOT execute.
 
-        NOTE: This method is depreciated and will be removed in the
-        next major release. Please see examples/minimqtt_pub_sub_blocking.py
-        for an example of creating a blocking loop which can handle wireless
-        network events.
-
+        .. note:: This method is depreciated and will be removed in the
```
brentru, line 695: > I noticed there was a (un-admonished) deprecation notice and a reference to an example that was not being displayed on RTD.

In the next PR (with the `fixed_header` change), could you also remove this method?

2bndy5: Yes; can do. Prepping for the next "major" release? No planned "minor" releases in immediate future? Is the [Bluetooth PR](#15 ) such a significant change? I also see you have a branch for compatibility with CPython sockets...

adafruit_minimqtt/adafruit_minimqtt.py
2bndy5: ah, you caught me... good eye. Opened a new PR with this change

adafruit_minimqtt/adafruit_minimqtt.py
2bndy5: Personally I'm not a fan of the black linter but request granted.

adafruit_minimqtt/adafruit_minimqtt.py
2bndy5: Opened #45 to satisfy request

brentru: @2bnd5 re: I'm not ready to merge the CPython sockets branch just yet, but yes, that'll be a major release.

PR #46, opened 2020-07-22
adjust fixed_header init; (-) loop_forever()

2bndy5:
@brentru This should now be rebased upon the adafruit master branch.
changes include:
* removed deprecated `loop_forever()` in preparation for next major release.
* consolidated the `fixed_header` initialization to 1 line of code.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -460,13 +460,11 @@ class MQTT:
             0 <= qos <= 1
         ), "Quality of Service Level 2 is unsupported by this library."
 
-        pub_hdr_fixed = bytearray()  # fixed header
-        pub_hdr_fixed.extend(MQTT_PUB)
-        pub_hdr_fixed[0] |= retain | qos << 1  # [3.3.1.2], [3.3.1.3]
+        # fixed header. [3.3.1.2], [3.3.1.3]
+        pub_hdr_fixed = bytearray([MQTT_PUB[0] | retain | qos << 1])
 
-        pub_hdr_var = bytearray()  # variable header
-        pub_hdr_var.append(len(topic) >> 8)  # Topic length, MSB
-        pub_hdr_var.append(len(topic) & 0xFF)  # Topic length, LSB
+        # variable header = 2-byte Topic length (big endian)
+        pub_hdr_var = struct.pack('>H', len(topic))
         pub_hdr_var.extend(topic.encode("utf-8"))  # Topic name
 
         remaining_length = 2 + len(msg) + len(topic)
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -303,8 +303,7 @@ class MQTT:
                 raise MMQTTException("Invalid broker address defined.", e)
 
         # Fixed Header
-        fixed_header = bytearray()
-        fixed_header.append(0x10)
+        fixed_header = bytearray([0x10])
 
         # NOTE: Variable header is
         # MQTT_HDR_CONNECT = bytearray(b"\x04MQTT\x04\x02\0\0")
@@ -687,21 +686,6 @@ class MQTT:
             while subscribed_topics:
                 feed = subscribed_topics.pop()
                 self.subscribe(feed)
-
-    def loop_forever(self):
-        """Starts a blocking message loop. Use this
-        method if you want to run a program forever.
-        Code below a call to this method will NOT execute.
-
-        .. note:: This method is depreciated and will be removed in the
-            next major release. Please see
-            `examples/minimqtt_pub_sub_blocking.py <examples.html#basic-forever-loop>`_
-            for an example of creating a blocking loop which can handle wireless
-            network events.
-        """
-        while True:
-            if self._sock.connected:
-                self.loop()
 
     def loop(self):
         """Non-blocking message loop. Use this method to
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -464,7 +464,7 @@ class MQTT:
         pub_hdr_fixed = bytearray([MQTT_PUB[0] | retain | qos << 1])
 
         # variable header = 2-byte Topic length (big endian)
-        pub_hdr_var = struct.pack('>H', len(topic))
+        pub_hdr_var = struct.pack(">H", len(topic))
         pub_hdr_var.extend(topic.encode("utf-8"))  # Topic name
 
         remaining_length = 2 + len(msg) + len(topic)
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -464,7 +464,7 @@ class MQTT:
 
         # variable header = 2-byte Topic length (big endian)
         pub_hdr_var = bytearray(struct.pack(">H", len(topic)))
-        pub_hdr_var.append(topic.encode("utf-8"))  # Topic name
+        pub_hdr_var.extend(topic.encode("utf-8"))  # Topic name
 
         remaining_length = 2 + len(msg) + len(topic)
         if qos > 0:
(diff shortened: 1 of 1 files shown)


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -303,8 +303,7 @@ def connect(self, clean_session=True):
                 raise MMQTTException("Invalid broker address defined.", e)
 
         # Fixed Header
-        fixed_header = bytearray()
-        fixed_header.append(0x10)
+        fixed_header = bytearray([0x10])
```
brentru, line 306: Could you also do the same to the fixed header packet within `publish`: https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/master/adafruit_minimqtt/adafruit_minimqtt.py#L464? Thanks!

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -688,21 +687,6 @@ def reconnect(self, resub_topics=True):
                 feed = subscribed_topics.pop()
                 self.subscribe(feed)
 
-    def loop_forever(self):
```
brentru, line 691: Self-note that once this PR is approved and tested, we'll need to file an issue in the following libraries and remove the deprecated `loop_forever` method:

CircuitPython Adafruit IO

- [ ] https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/blob/master/adafruit_io/adafruit_io.py#L231

CircuitPython GCP IoT Core

- [ ] https://github.com/adafruit/Adafruit_CircuitPython_GC_IOT_Core/blob/master/adafruit_gc_iot_core.py#L222

adafruit_minimqtt/adafruit_minimqtt.py
2bndy5: Again, good eye! I'd like to note that the `MQTT_PUB` const is already a bytearray & only used once throughout `adafruit_minimqtt.py`. Now my initial reaction is to get rid of the (rather non-descriptive) const and change the line on L464 to something like `pub_hdr_fixed = bytearray([0x30 | retain | qos << 1])`. This would be less computationally intensive as opposed to `pub_hdr_fixed = bytearray([MQTT_PUB[0] | retain | qos << 1])`. If you approve the first solution, then I'll go with the first solution. Otherwise I'll just commit the second solution.

PS not sure what the comment `# [3.3.1.2], [3.3.1.3]` is about on L466. I'm assuming its a reference to some revision of a MQTT standard that I'm unaware of.

2bndy5: reading the source further, I think I can consolidate the [lines 468-470](https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/1adb75d1f247e740df07b62d40a23e0dfdc6cb5d/adafruit_minimqtt/adafruit_minimqtt.py#L468-L470) to  `pub_hdr_var = struct.pack('>H', len(topic))` seen as how `struct` is already being imported.

brentru: > PS not sure what the comment # [3.3.1.2], [3.3.1.3] is about on L466. I'm assuming its a reference to some revision of a MQTT standard that I'm unaware of.

http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718039

>  Now my initial reaction is to get rid of the const and change the line on L464 to something like pub_hdr_fixed = bytearray([0x30 | retain | qos << 1])

I like this solution, let's stick with it.

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -461,13 +460,11 @@ def publish(self, topic, msg, retain=False, qos=0):
...
-        pub_hdr_fixed.extend(MQTT_PUB)
-        pub_hdr_fixed[0] |= retain | qos << 1  # [3.3.1.2], [3.3.1.3]
+        # fixed header. [3.3.1.2], [3.3.1.3]
+        pub_hdr_fixed = bytearray([MQTT_PUB[0] | retain | qos << 1])
 
-        pub_hdr_var = bytearray()  # variable header
-        pub_hdr_var.append(len(topic) >> 8)  # Topic length, MSB
-        pub_hdr_var.append(len(topic) & 0xFF)  # Topic length, LSB
+        # variable header = 2-byte Topic length (big endian)
+        pub_hdr_var = struct.pack(">H", len(topic))
```
brentru, line 467: Running the example `simpletest` on `Adafruit CircuitPython 5.3.1 on 2020-07-13; Adafruit PyPortal with samd51j20` throws the following error in the REPL:

```
Subscribed to test/topic with QOS level 0
Publishing to test/topic
Traceback (most recent call last):
  File "code.py", line 116, in <module>
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 468, in publish
AttributeError: 'bytes' object has no attribute 'extend'
```

Instead of `extend` on L468, we may want to `append`

2bndy5: I feel like this is a borderline rookie mistake. I tried `append()` instead, but I still got the same error `AttributeError: 'bytes' object has no attribute 'append'` However `pub_hdr_var += topic.encode()` does what we want. The real problem comes from the change to using `struct.pack()` which returns an immutable `bytes` object, where methods like `append()` or `extend(bytes([0x00]))` (notice the need for conversion from `int` to `bytes` when using `extend()`) can only apply to a mutable `bytearray` object. So, I've wrapped the `struct.pack()` into a `bytearray(struct.pack())` which allows us to keep using the many `pub_hdr_var.append()` calls after L468.

adafruit_minimqtt/adafruit_minimqtt.py
2bndy5: thanks for the reply. I've commit the first solution. Also, thanks for the link to MQTT specifications (primary reason I like to contribute is for opportunities to learn).

adafruit_minimqtt/adafruit_minimqtt.py
2bndy5: these issue have been opened per request.
See [issue for Adafruit_CircuitPython_AdafruitIO ](https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/issues/40) & 
[issue for Adafruit_CircuitPython_GC_IOT_Core ](https://github.com/adafruit/Adafruit_CircuitPython_GC_IOT_Core/issues/13)

brentru: Wow, thanks!

brentru:
@2bndy5  Do you want to remove the calls to `loop_forever` within those two libs? Do you want me to?

2bndy5:
I can open the PR for each since you did all the legwork in testing this in the hardware. Literally the least I could do.

2bndy5:
@brentru [PR for Adafruit_CircuitPython_GC_IOT_Core](https://github.com/adafruit/Adafruit_CircuitPython_GC_IOT_Core/pull/14) & [PR for Adafruit_CircuitPython_AdafruitIO](https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/pull/41) are open with all checks passing.

brentru:
Thanks for opening the PRs. Looks like examples are working now. Will merge once I test CircuitPython IO.

Tested on `Adafruit CircuitPython 5.3.1 on 2020-07-13; Adafruit PyPortal with samd51j20`
```

Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.
code.py output:
Connecting to WiFi...
Connected!
Connecting to Adafruit IO...
Connected to Adafruit IO! Listening for topic changes on brubell/feeds/onoff
Sending photocell value: 0...
Sent!
Sending photocell value: 1...
Sent!
Sending photocell value: 2...
Sent!
Sending photocell value: 3...
Sent!
Sending photocell value: 4...
Sent!
Sending photocell value: 5...
Sent!
Sending photocell value: 6...
Sent!
Sending photocell value: 7...
Sent!
Sending photocell value: 8...
Sent!
Sending photocell value: 9...
Sent!
Sending photocell value: 10...
Sent!
Sending photocell value: 11...
Sent!
Sending photocell value: 12...
Sent!
Sending photocell value: 13...
Sent!
Sending photocell value: 14...
Sent!
Sending photocell value: 15...
Sent!
Sending photocell value: 16...
Sent!
Sending photocell value: 17...
Sent!
Sending photocell value: 18...
Sent!
Sending photocell value: 19...
Sent!
Sending photocell value: 20...
Sent!
Sending photocell value: 21...
Sent!
Sending photocell value: 22...
Sent!```

Issue #39, opened 2020-06-07
The default value for the payload parameter is not supported in the will_set method 


jimbobbennett:
To reproduce:

Call the `will_set` method and leave the `payload` parameter as its default value of `None`.

This causes an exception of `Message can not be none`.

If the payload can't be `None`, then it should not have a default value of `None` - all default values should work as their existence implies that the value is allowed.

2bndy5:
To be clear, are we expecting to mimic paho_mqtt behavior? In that case, a payload of `None` would be replaced with a zero-length string as described in `will_set()` of the [paho_mqtt docs](https://pypi.org/project/paho-mqtt/#option-functions).

brentru:
@2bndy5  @jimbobbennett  Yes, I'd like to be in-like with paho mqtt as much as possible in this library.

brentru:
fixed by https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/pull/47

PR #47, opened 2020-07-23
payload parameter of type `None` replaced with zero-length string in `will_set()`

2bndy5:
satisfies #39 in accordance with paho_mqtt library compatibility

I tested this fix locally (on my trusty ItsyBitsy M4) without connection to a broker. The only esp32 I have lying around is a TinyPico, so properly testing this on my end might be a bit taxing for a 1-line fix. If there's an easier way to test this with a Raspberry Pi, I'd gladly dust off my old RPi.

I followed the data from the `payload` parameter through the code to see if this fix instigates another issue, and I found no problems. However, I'm not sure about `self._sock.send(string)` where `self._lw_msg` (instance attribute that holds the zero-length string when payload parameter is of type `None`) is passed as the `string` parameter in [`_send_str()`](https://github.com/2bndy5/Adafruit_CircuitPython_MiniMQTT/blob/95f9519333d53a426b89c7194d1f843b4863f1fd/adafruit_minimqtt/adafruit_minimqtt.py#L778). BTW, I have verified that `self._lw_msg` becomes `b''` (empty bytearray) when calling `will_set()` using parameters' defaults with this fix. 

### TL;DR 
All this to say that the success of this fix depends on the socket interface's ability to send an empty bytearray.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -217,7 +217,7 @@ class MQTT:
         if self._is_connected:
             raise MMQTTException("Last Will should only be called before connect().")
         if payload is None:
-            raise MMQTTException("Message can not be None.")
+            payload = ""
         if isinstance(payload, (int, float, str)):
             payload = str(payload).encode()
         else:


2bndy5:
Great news! Does this mean that the #39 can be closed?

Issue #54, opened 2021-01-12
Wrong semantics for size in socket.recv() causes sporatic failures in message parsing


flavio-fernandes:
The code for socket read has a buffer size param that indicates the _maximum_ number of bytes
that can be returned within a single call:

https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/fce466bd2bb70ca86b79e5cb36bbaca00afacfd1/adafruit_esp32spi/adafruit_esp32spi_socket.py#L107-L112

Trouble is, MQTT is using that param as if that would be the _minimum_ number of bytes that should be returned. 

Also, ping method tries to drain buffer looking for the PINGRESP message and if that is not the message in the
receive buffer, it never reverts the 1 byte it took out of the buffer.

https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/682de9627ed54016463e39c3660650184d9b9fda/adafruit_minimqtt/adafruit_minimqtt.py#L406-L412

After that point, MQTT broker and client are 'out of sync', causing any future exchanges to be 'out of whack'.
A better approach would be to flag that a ping response is expected, but handle it together with any other MQTT
message via _wait_for_msg

```
1106.57: DEBUG - KeepAlive period elapsed - requesting a PINGRESP from the server...
1106.57: DEBUG - Sending PINGREQ
Writing: b'\xc0\x00'
1106.63: DEBUG - Checking PINGRESP
ESPSocket: 27 bytes available
Reading 1 bytes from ESP socket with status 4
ESPSocket: 26 bytes available
Reading 1 bytes from ESP socket with status 4
ESPSocket: 25 bytes available
Reading 2 bytes from ESP socket with status 4
ESPSocket: 23 bytes available
Reading 21 bytes from ESP socket with status 4
ESPSocket: 2 bytes available
Reading 2 bytes from ESP socket with status 4
Traceback (most recent call last):
  File "code.py", line 297, in <module>
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 704, in loop
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 407, in ping
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 734, in _wait_for_msg
UnicodeError: 
```

PR #55, opened 2021-01-13
Fix socket receive issues related to message buffer size

flavio-fernandes:
This change addresses a few issues in the handling of the MQTT
messages that caused the library to become unstable:

- Add wapper for socket.recv() so that an exact number of bytes
  are read into the buffer before attempting to parse the MQTT
  message;

- Fix handling of ping response packets via _wait_for_msg();

- Fix disconnect so it can gracefully handle cases when socket writes
  are not possible. Also re-init _subscribed_topics as an empty list
  instead of None.

Related-to https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/issues/102
Fixes https://github.com/adafruit/Adafruit_CircuitPython_PyPortal/issues/98
Fixes https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/issues/54
Signed-off-by: Flavio Fernandes <flavio@flaviof.com>

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -175,6 +175,35 @@ class MQTT:
     def __exit__(self, exception_type, exception_value, traceback):
         self.deinit()
 
+    def _sock_exact_recv(self, bufsize):
+        """Reads _exact_ number of bytes from the connected socket. Will only return
+        string with the exact number of bytes requested.
+
+        The semantics of native socket receive is that it returns no more than the
+        specified number of bytes (i.e. max size). However, it makes no guarantees in
+        terms of the minimum size of the buffer, which could be 1 byte. This is a
+        wrapper for socket recv() to ensure that no less than the expected number of
+        bytes is returned or trigger a timeout exception.
+
+        :param int bufsize: number of bytes to receive
+        """
+        stamp = time.monotonic()
+        rc = self._sock.recv(bufsize)
+        to_read = bufsize - len(rc)
+        assert to_read >= 0
+        read_timeout = self.keep_alive
+        while to_read > 0:
+            recv = self._sock.recv(to_read)
+            to_read -= len(recv)
+            rc += recv
+            if time.monotonic() - stamp > read_timeout:
+                raise MMQTTException(
+                    "Unable to receive {} bytes within {} seconds.".format(
+                        to_read, read_timeout
+                    )
+                )
+        return rc
+
     def deinit(self):
         """De-initializes the MQTT client and disconnects from the mqtt broker."""
         self.disconnect()
@@ -351,7 +380,7 @@ class MQTT:
         while True:
             op = self._wait_for_msg()
             if op == 32:
-                rc = self._sock.recv(3)
+                rc = self._sock_exact_recv(3)
                 assert rc[0] == 0x02
                 if rc[2] != 0x00:
                     raise MMQTTException(CONNACK_ERRORS[rc[2]])
@@ -366,32 +395,38 @@ class MQTT:
         self.is_connected()
         if self.logger is not None:
             self.logger.debug("Sending DISCONNECT packet to broker")
-        self._sock.send(MQTT_DISCONNECT)
+        try:
+            self._sock.send(MQTT_DISCONNECT)
+        except RuntimeError as e:
+            if self.logger:
+                self.logger.warning("Unable to send DISCONNECT packet: {}".format(e))
         if self.logger is not None:
             self.logger.debug("Closing socket")
         self._sock.close()
         self._is_connected = False
-        self._subscribed_topics = None
+        self._subscribed_topics = []
         if self.on_disconnect is not None:
             self.on_disconnect(self, self.user_data, 0)
 
     def ping(self):
         """Pings the MQTT Broker to confirm if the broker is alive or if
         there is an active network connection.
+        Returns response codes of any messages received while waiting for PINGRESP.
         """
         self.is_connected()
-        if self.logger is not None:
+        if self.logger:
             self.logger.debug("Sending PINGREQ")
         self._sock.send(MQTT_PINGREQ)
-        if self.logger is not None:
-            self.logger.debug("Checking PINGRESP")
-        while True:
-            op = self._wait_for_msg(0.5)
-            if op == 208:
-                ping_resp = self._sock.recv(2)
-                if ping_resp[0] != 0x00:
-                    raise MMQTTException("PINGRESP not returned from broker.")
-            return
+        ping_timeout = self.keep_alive
+        stamp = time.monotonic()
+        rc, rcs = None, []
+        while rc != MQTT_PINGRESP:
+            rc = self._wait_for_msg()
+            if rc:
+                rcs.append(rc)
+            if time.monotonic() - stamp > ping_timeout:
+                raise MMQTTException("PINGRESP not returned from broker.")
+        return rcs
 
     # pylint: disable=too-many-branches, too-many-statements
     def publish(self, topic, msg, retain=False, qos=0):
@@ -486,9 +521,9 @@ class MQTT:
             while True:
                 op = self._wait_for_msg()
                 if op == 0x40:
-                    sz = self._sock.recv(1)
+                    sz = self._sock_exact_recv(1)
                     assert sz == b"\x02"
-                    rcv_pid = self._sock.recv(2)
+                    rcv_pid = self._sock_exact_recv(2)
                     rcv_pid = rcv_pid[0] << 0x08 | rcv_pid[1]
                     if pid == rcv_pid:
                         if self.on_publish is not None:
@@ -571,7 +606,7 @@ class MQTT:
         while True:
             op = self._wait_for_msg()
             if op == 0x90:
-                rc = self._sock.recv(4)
+                rc = self._sock_exact_recv(4)
                 assert rc[1] == packet[2] and rc[2] == packet[3]
                 if rc[3] == 0x80:
                     raise MMQTTException("SUBACK Failure!")
@@ -634,7 +669,7 @@ class MQTT:
         while True:
             op = self._wait_for_msg()
             if op == 176:
-                return_code = self._sock.recv(3)
+                return_code = self._sock_exact_recv(3)
                 assert return_code[0] == 0x02
                 # [MQTT-3.32]
                 assert (
@@ -671,6 +706,7 @@ class MQTT:
     def loop(self):
         """Non-blocking message loop. Use this method to
         check incoming subscription messages.
+        Returns response codes of any messages received.
         """
         if self._timestamp == 0:
             self._timestamp = time.monotonic()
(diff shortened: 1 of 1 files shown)


flavio-fernandes:
Hi @tannewt . I could use some much needed help on the CI failure, as well as code review from you on this PR.
I believe this addresses the issue we were talking about in https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/issues/102

ladyada:
https://learn.adafruit.com/contribute-to-circuitpython-with-git-and-github has some info about running black/pylint
@dherrada can you help with CI passing Q's

flavio-fernandes:
> https://learn.adafruit.com/contribute-to-circuitpython-with-git-and-github has some info about running black/pylint
> @dherrada can you help with CI passing Q's

Thank you for the link. It seems that Travis is no longer used and https://learn.adafruit.com/contribute-to-circuitpython-with-git-and-github/open-pull-request needs an update, agree? It would be nice to add more info on running black tool
on that page too.

@dherrada : The ci issues are not related to my changes so I will definitely need some help. I managed to get past the
lint, but have no idea on how to tackle this one:

```
Unfortunately, your project is not compliant with version 3.0 of the REUSE Specification :-(
```

Best,

-- flaviof

flavio-fernandes:
Hi @brentru ! Based on your involvement with https://learn.adafruit.com/pyportal-weather-station , I would like very
much if you tried / reviewed the changes I'm proposing on this PR. Could you, pretty please? ;)

adafruit_minimqtt/matcher.py
```diff
@@ -1,100 +1,99 @@
-# Copyright (c) 2017 Yoch <https://github.com/yoch>
```
brentru, line 1: Please put `matcher.py` back, thanks

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -192,6 +193,35 @@ def __enter__(self):
     def __exit__(self, exception_type, exception_value, traceback):
         self.deinit()
 
+    def _sock_exact_recv(self, bufsize):
```
brentru, line 196: Instead of adding `socket_exact_recv`, could you instead modify `sock_recv`?

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -379,17 +409,20 @@ def connect(self, clean_session=True):
...
 
     def disconnect(self):
-        """Disconnects the MiniMQTT client from the MQTT broker.
-        """
+        """Disconnects the MiniMQTT client from the MQTT broker."""
         self.is_connected()
         if self.logger is not None:
             self.logger.debug("Sending DISCONNECT packet to broker")
-        self._sock.send(MQTT_DISCONNECT)
+        try:
```
brentru, line 416: This try/retry does not make sense to me - Why would we not be able to send a DISCONNECT control packet?

Disconnect packets (https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718094) expect no response payload and should only close the connection with the server (and in our case, empty a list of prvsly subscribed topics)

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -379,17 +409,20 @@ def connect(self, clean_session=True):
...
+            self._sock.send(MQTT_DISCONNECT)
+        except RuntimeError as e:
+            if self.logger:
+                self.logger.warning("Unable to send DISCONNECT packet: {}".format(e))
         if self.logger is not None:
             self.logger.debug("Closing socket")
         self._sock.close()
         self._is_connected = False
-        self._subscribed_topics = None
+        self._subscribed_topics = []
```
brentru, line 425: May want to check for `len(_subscribed_topics)` first, then empty it?

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -398,18 +431,15 @@ def ping(self):
         there is an active network connection.
         """
         self.is_connected()
+        if self._pending_ping_response:
+            self._pending_ping_response = False
+            raise MMQTTException("Ping response was pending from previous MQTT_PINGREQ")
```
brentru, line 436: Don't raise an exception here, raise one within the method which calls `ping()`

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -398,18 +431,15 @@ def ping(self):
         there is an active network connection.
         """
         self.is_connected()
+        if self._pending_ping_response:
+            self._pending_ping_response = False
```
brentru, line 435: Instead of handling the polling for `PINGRESP` within `wait_for_msg`, we should be handling it within this method. 

Otherwise, any method calling `ping`, such as user-code within `code.py` will need to implement code to poll for a pingresp.

> Also, ping method tries to drain buffer looking for the PINGRESP message and if that is not the message in the
receive buffer, it never reverts the 1 byte it took out of the buffer. via https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/issues/54

Could we implement this within `ping()` instead?

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -709,27 +739,37 @@ def _wait_for_msg(self, timeout=30):
         """Reads and processes network events.
         Returns response code if successful.
         """
+        # Check if there are any messages by reading 1 byte from socket
+        # This is a place where it is okay to have 'less' than what we asked.
```
brentru, line 743: Please remove comment

brentru (changes requested):

dherrada:
@flavio-fernandes master is now passing, but there are a few conflicts. I would recommend running pre-commit locally to make the conflicts smaller and then resolving the remaining ones manually

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: Heh, that is a question for @tannewt :) 
Based on [this thread](https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/issues/102#issuecomment-754947339), I got the impression he did not want the fix in Adafruit_CircuitPython_ESP32SPI. I could be wrong.

adafruit_minimqtt/matcher.py
flavio-fernandes: will do

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: That is because sock.write() invoked by disconnect may fail. That happens if tcp session is in a bad state or wifi is down.

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: Maybe so, but what is important is that we have a list here. 
Otherwise we will get something like this when trying to connect later on:
```
  File "code.py", line 287, in <module>
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 406, in connect
  File "code.py", line 137, in connect
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 628, in subscribe
AttributeError: 'NoneType' object has no attribute 'append'

Press any key to enter the REPL. Use CTRL-D to reload.
```
Do you mind making that shange as part of another PR?

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: why?

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: So that is the thing. Imagine that just before we send this ping a message that is not "pingresp" arrives. We would be duplicating wait_for_msg in order to fish the pingresp out of the list of unrelated messages. Having a flag to indicate whether the ping arrived is a cleaner way of checking for the response, don't you thinkk?

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: ok. will do.

flavio-fernandes:
> @flavio-fernandes master is now passing, but there are a few conflicts. I would recommend running pre-commit locally to make the conflicts smaller and then resolving the remaining ones manually

Sounds good. How do you run the pre-commit locally? All I got so far is "pip install black ; black file.py" ;)

adafruit_minimqtt/adafruit_minimqtt.py
tannewt: Yup, socket recv doesn't wait for all of the bytes. This is the right place to do it. Changing recv would break working with CPython's socket.

dherrada:
> How do you run the pre-commit locally?
@flavio-fernandes Install pre-commit with pip, and then just run it with `pre-commit run --all-files`. Note that you should copy this (https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/master/.pre-commit-config.yaml) file over to your fork to get it to run right.

adafruit_minimqtt/adafruit_minimqtt.py
brentru: > Do you mind making that change as part of another PR?

I think it should be folded into this PR since we're introducing a list here and want to handle the edge case.

adafruit_minimqtt/adafruit_minimqtt.py
brentru: ok!

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: > > Do you mind making that change as part of another PR?
> 
> I think it should be folded into this PR since we're introducing a list here and want to handle the edge case.

Oh... that is not new. If you look at:  https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/e02d658425f099c5a573cd0e0c2b3b4150c25cfd/adafruit_minimqtt/adafruit_minimqtt.py#L162  before this PR, you can see that this was always a list. The disconnect code is the 'bug' that was improperly setting it to None instead of [].

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: One thing we could do is to add something like:

```
other_msgs = []
while not self._pending_ping_response:
    msg = self.wait_for_msg()
    if msg:
        other_msgs.append(msg)
    if <time_to_give_up>:
        raise <ping response did not make it...>

return other_msgs
```

Would that be better?

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: @brentru if you would like, I can undo this change on this PR and open a separate issue for specifically dealing with the disconnect() / connect() bug. LMK, pls.

adafruit_minimqtt/adafruit_minimqtt.py
flavio-fernandes: @brentru if you would like, I can undo this change on this PR and open a separate issue for specifically dealing with the disconnect() / connect() bug. LMK, pls.

adafruit_minimqtt/adafruit_minimqtt.py
brentru: I understand what you're trying to do here and agree that `socket.send` could fail if there was a network communication issue. Let's keep it in this PR 👍

brentru:
@flavio-fernandes  This is passing, do you want me to re-review on hardware?

brentru (approved):
@flavio-fernandes  Thanks for making the requested changes! Tested simpletest and adafruit io on `Adafruit CircuitPython 6.1.0-rc.1 on 2021-01-15; Adafruit PyPortal with samd51j20`

Issue #58, opened 2021-01-27
Dev Branch 'SocketPool' object has no attribute 'timeout'


makermelissa:
I found this on the ESP32-S2 Metro using the Dev Branch. It only happens occasionally"
```
Traceback (most recent call last):
  File "code.py", line 140, in <module>
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 862, in loop
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 874, in _wait_for_msg
AttributeError: 'SocketPool' object has no attribute 'timeout'
```

makermelissa:
I replaced line 874 with `except OSError:` and that seems to be working now.

brentru:
@makermelissa  Thanks for finding this type of bug. Which `OSError` does it throw?

makermelissa:
That's the thing, it never does throw an error. It just didn't like that timeout didn't exist in SocketPool, which is why I noted that it's probably not the right way to do it.

makermelissa:
Ok, removed the try/except bloack and got:
```
OSError: [Errno 116] ETIMEDOUT
```

brentru:
Thanks for finding it, I'll catch that specific error

makermelissa:
Ok this code block seems pretty solid:
```
        try:
            self._sock.recv_into(res, 1)
            print("Resp: ", res) # TODO BR: Remove debugging
        except OSError as error:
            if error.errno == errno.ETIMEDOUT:
                return None
            else:
                raise
```

brentru:
This has been addressed in #61 and released.

Issue #43, opened 2020-07-10
CPython socket compatibility


brentru:

brentru:
I'm working on this over on https://github.com/brentru/Adafruit_CircuitPython_MiniMQTT/tree/update-for-cpython-compat

askpatrickw:
Hey @brentru, I was trying to use your library with 
https://github.com/tannewt/Adafruit_CircuitPython_Requests/blob/http11/adafruit_requests.py#L358
And I ran into an issue of the socket_module not being set.

I tried to mimic what _get_socket in requests is doing there by adding socket_pool and ssl_context parameters to MQTT.__init__ and commenting out the check for socket_module in _get_socket. But that lead to me getting an error in connect.

```
Traceback (most recent call last):
  File "code.py", line 80, in <module>
  File "/Users/askpatrickw/Downloads/Adafruit_CircuitPython_MiniMQTT-update-for-cpython-compat/adafruit_minimqtt/adafruit_minimqtt.py", line 306, in connect
TypeError: function missing required positional argument #3
``` 

I'm tired but I'll probably have a go at this in a day or so. If  you had any obvious feedback let me know. I'm happy to push into your fork if I get it working.

I should mention this is on Scott's ESP32-S2 Native_wifi branch and with his requests module I linked above.

Cheers!

brentru:
@askpatrickw  I'm not sure what you're using the Requests module for, this library doesn't rely on it.

Could you post your full _code.py_ using this branch so I can take a look at it? I have not tested this on ESP32-S2 yet - waiting on scott's requests fork to be PR'd so I can keep this branch in-line with the final implementation.

askpatrickw:
Hey @brentru thanks for offering to take a look, here's a full walk through first with your code and then with my attempts...

My code.py is essentially the simple example I'll only show the setup and connect. 

There is no way to do his part:
`MQTT.set_socket(socket, esp)` so I commented it out.

``` python
client = MQTT.MQTT(
    broker=secrets["broker"], username=secrets["user"], password=secrets["pass"],
)

# Connect callback handlers to client
client.on_connect = connect
client.on_disconnect = disconnect
client.on_subscribe = subscribe
client.on_unsubscribe = unsubscribe
client.on_publish = publish

print("Attempting to connect to %s" % client.broker)
client.connect()
```
Output
```python
Attempting to connect to io.adafruit.com
Traceback (most recent call last):
[...]

PR #61, opened 2021-02-01
Add ESP32-S2 (Native Networking) and CPython 

brentru:
This pull request:
* Adds compatibility and examples for ESP32-S2 and "Native Networking"
* Adds compatibility and examples for usage with CPython
  * **NOTE:** CPython SSL Sockets **are not** currently supported. They will be supported in a subsequent release. SSL sockets have different behavior from regular CPython sockets (https://docs.python.org/3/library/ssl.html#notes-on-non-blocking-sockets)
* Provides backwards-compatible "legacy" support for ESP32SPI, Ethernet socket interface. 
* Removes dependency for Adafruit_CircuitPython_Logging to be installed, a logger may be passed into a MQTT object.
* Adds a `username_pw_set` method to dynamically set username and password outside of client initialization.
* Reorganizes examples into folders, categorized by network interface 
* `connect()` may now accept a keepalive time
* `loop()` may now accept a `timeout` in seconds.

**NOTE: "Native Socket" (ESP32-S2) is only supported in CircuitPython 6.2.0b1 or _newer_**

Addresses:
https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/issues/58
https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/issues/43
https://github.com/brentru/Adafruit_CircuitPython_MiniMQTT/pull/2
https://github.com/adafruit/Adafruit_CircuitPython_AzureIoT/issues/21
https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/issues/52

This library has been tested on:
`Adafruit CircuitPython 6.2.0-beta.1 on 2021-01-27; Adafruit Metro ESP32S2 with ESP32S2`, **Native Networking**
`Adafruit CircuitPython 6.0.1 on 2020-12-28; Adafruit Feather M4 Express with samd51j19` with **Ethernet** FeatherWing
`Adafruit CircuitPython 6.1.0 on 2021-01-21; Adafruit PyPortal with samd51j20`, **ESP32SPI**
**CPython 3.9.1**


makermelissa (approved):
Tested and it works great. That's a bonus that it doesn't require logging anymore.

brentru:
**Note**: The TLS implementation for native sockets is moving into a new class, `SSLSocket` in https://github.com/adafruit/circuitpython/pull/4049. We will need to update how MiniMQTT creates/handles SSL sockets once that code is pulled in.

brentru:
Thanks for reviewing and testing

Issue #62, opened 2021-02-03
CPython example not working with secure broker


brentru:
This library currently works with CPython sockets connecting to a insecure MQTT broker port (1883) but falls down when attempting to connect to a secure MQTT broker port (8883):

```Traceback (most recent call last):
  File "/Users/brentrubell/Desktop/github_brentru/Adafruit_CircuitPython_MiniMQTT/test_cpython.py", line 63, in <module>
    mqtt_client.connect()
  File "/Users/brentrubell/Desktop/github_brentru/Adafruit_CircuitPython_MiniMQTT/adafruit_minimqtt/adafruit_minimqtt.py", line 467, in connect
    self._sock = self._get_socket(self.broker, self.port)
  File "/Users/brentrubell/Desktop/github_brentru/Adafruit_CircuitPython_MiniMQTT/adafruit_minimqtt/adafruit_minimqtt.py", line 284, in _get_socket
    sock = self._ssl_context.wrap_socket(sock, server_hostname=host)
  File "/Users/brentrubell/Desktop/github_brentru/Adafruit_CircuitPython_MiniMQTT/adafruit_minimqtt/adafruit_minimqtt.py", line 114, in wrap_socket
    return _FakeSSLSocket(socket, self._iface.TLS_MODE)
AttributeError: 'NoneType' object has no attribute 'TLS_MODE'```

**Comment by brentru** [comment]
Digging deeper, this is more an issue with the example itself and the macOS SSL installation under Python3 than the library, I'll be making a pull request shortly to fix it.

In the future, if a user on macOS gets `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1123)`, they'll need to upgrade the `certifi` package by running the following command: 

`/Applications/Python\ 3.9/Install\ Certificates.command`

**Comment by tannewt** [comment]
Why are you using FakeSSLSocket when using CPython? Your self._ssl_context should be from the real `ssl` library.

**Comment by tannewt** [comment]
FakeSSLSocket should only be used with ESP32SPI.

**Comment by brentru** [comment]
@tannewt  The code which threw this error incorrectly did not pass the `ssl_context` into the MiniMQTT constructor and looked like:

```
# Set up a MiniMQTT Client
mqtt_client = MQTT.MQTT(
    broker=secrets["broker"],
    username=secrets["aio_username"],
    password=secrets["aio_key"],
[...]

PR #65, opened 2021-02-04
Fix CPython/Legacy Socket SSL Problems

brentru:
This pull request:
* Sets `_FakeSSLContext` within calls to `set_interface` instead of within `__init__`
* Removes socket reuse code mirrored in from `adafruit-circuitpython-requests` 
* Calls to `reconnect` now close an open socket
* Legacy socket API set up in `_get_connect_socket` instead of within `__init__`
* Updated `examples/cpython/minimqtt_simpletest_cpython` to address #62 

Addresses https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/issues/62

Tested on: ESP32SPI (PyPortal), CPython, and ESP32-S2 (Metro ESP32-S2)


tannewt (approved):
Looks fine to me! I didn't test it.

PR #79, opened 2021-04-02
MQTT QOS=1,2 rules state that pid must be > 0.

dlizotte-uwo:
Publishing with QOS 1 seemed to be broken because PID was incremented immediately *after* putting into packet, meaning that the PUBACK packet would not match the new (incremented) PID.

There is a danger if publishing more than 0xFFFF packets that .to_bytes will fail because it can't pack into 2 bytes after that. PID must wrap around to 1, not 0, according to spec, which demands that PID be nonzero for QOS 1 and 2.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -591,9 +591,9 @@ class MQTT:
         if qos > 0:
             # packet identifier where QoS level is 1 or 2. [3.3.2.2]
             remaining_length += 2
-            pub_hdr_var.append(0x00)
-            pub_hdr_var.append(self._pid)
-            self._pid += 1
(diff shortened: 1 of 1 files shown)


brentru:
@dlizotte-uwo  I tested and verified on wireshark that the packet identifier for pub/sub is => 1 when qos >1, also checked fields for anything malformed. Looks good:
Capturing_from_USB_10_100_1000_LAN__en7


> There is a danger if publishing more than 0xFFFF packets that .to_bytes will fail because it can't pack into 2 bytes after that. PID must wrap around to 1, not 0, according to spec, which demands that PID be nonzero for QOS 1 and 2.

I'm ok with wrapping to 1, thank you for handling this edge case.

brentru (approved):

Issue #86, opened 2021-08-07
Blocking loop() function and issues with PINGREQ and timeout=0


ajlennon:
Hi all,

Thanks for this - am trying to integrate it into some code and having some problems.

One is that I think a reasonable use-case is to call poll() repeatedly in my main loop. 

The poll() code documentation states it is non-blocking (as the PAHO call is for example) but it actually blocks for 1s which is not trivial for an embedded system main loop. I suggest timeout should default to 0.

That said, if I set timeout to zero there is a problem as the poll() call will be called multiple times while a ping request is outstanding and this results in multiple ping requests being made. I see 3-5 at a time with DEBUG logging on. 

This is because self._timestamp is being reset after self.ping()

The solution seems to be pretty straightforward which is to reset self._timestamp() before sending the ping()

e.g.

```
def loop(self, timeout=1):
        """Non-blocking message loop. Use this method to
        check incoming subscription messages.
        Returns response codes of any messages received.
        :param int timeout: Socket timeout, in seconds.
        """
        if self._timestamp == 0:
            self._timestamp = time.monotonic()
        current_time = time.monotonic()
        if current_time - self._timestamp >= self.keep_alive:
            self._timestamp = 0
            # Handle KeepAlive by expecting a PINGREQ/PINGRESP from the server
            if self.logger is not None:
                self.logger.debug(
                    "KeepAlive period elapsed - requesting a PINGRESP from the server..."
                )
            rcs = self.ping()
            return rcs
        self._sock.settimeout(timeout)
        rc = self._wait_for_msg()
        return [rc] if rc else None
```

brentru:
Fixed in #87, closing...

PR #87, opened 2021-08-07
adafruit_minimqtt: Fix issue #86 with multiple poll() PINGREQs

ajlennon:
If I set timeout to zero there is a problem as the poll() call will be
called multiple times while a ping request is outstanding and this
results in multiple ping requests being made. I see 3-5 at a time with
DEBUG logging on.

This is because self._timestamp is being reset after self.ping()

The solution seems to be pretty straightforward which is to reset
self._timestamp() before sending the ping()

Signed-off-by: Alex J Lennon <ajlennon@dynamicdevices.co.uk>


brentru (approved):

ajlennon:
Any time :) Thanks for all your hard work on the libraries!

PR #97, opened 2022-01-07
feat: add support for binary messages

alvarowolfx:
I'm creating a basic client library to use with [Golioth](https://golioth.io) Platform. We recently added a new gateway to support MQTT devices and that allows us to add support to CircuitPython.

But while developing the client, we had issues with one of our services meant for Over the Air Updates, which serves not only for the main device firmware, but you can send any arbitrary artifact ( like images/bitmap in my testing here). Currently, this MQTT library tries to convert all messages to a Unicode string, which unfortunately fails for binary data. Reference to where it fails: https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/main/adafruit_minimqtt/adafruit_minimqtt.py#L843

So what I'm proposing here is to add an optional parameter called `use_binary`, that would allow users to use the MQTT client in binary mode and return the raw `bytearray` message. Them is up to the user to convert to a string as needed.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -126,6 +126,7 @@ class MQTT:
     :param int keep_alive: KeepAlive interval between the broker and the MiniMQTT client.
     :param socket socket_pool: A pool of socket resources available for the given radio.
     :param ssl_context: SSL context for long-lived SSL connections.
+    :param bool use_binary_mode: Messages are passed as bytearray instead of string to callbacks.
 
     """
 
@@ -141,12 +142,14 @@ class MQTT:
         keep_alive=60,
         socket_pool=None,
         ssl_context=None,
+        use_binary_mode=False,
     ):
 
         self._socket_pool = socket_pool
         self._ssl_context = ssl_context
         self._sock = None
         self._backwards_compatible_sock = False
+        self._use_binary_mode = use_binary_mode
 
         self.keep_alive = keep_alive
         self._user_data = None
@@ -839,8 +842,9 @@ class MQTT:
             pid = pid[0] << 0x08 | pid[1]
             sz -= 0x02
         # read message contents
-        msg = self._sock_exact_recv(sz)
-        self._handle_on_message(self, topic, str(msg, "utf-8"))
+        raw_msg = self._sock_exact_recv(sz)
+        msg = raw_msg if self._use_binary_mode else str(raw_msg, "utf-8")
+        self._handle_on_message(self, topic, msg)
         if res[0] & 0x06 == 0x02:
             pkt = bytearray(b"\x40\x02\0\0")
             struct.pack_into("!H", pkt, 2, pid)


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -126,6 +126,7 @@ class MQTT:
     :param int keep_alive: KeepAlive interval between the broker and the MiniMQTT client.
     :param socket socket_pool: A pool of socket resources available for the given radio.
     :param ssl_context: SSL context for long-lived SSL connections.
+    :param bool use_binary: Sets if message is passed as bytearray instead of string to callbacks
```
brentru, line 129: Possibly `use_binary_mode`? 

Are there other libraries that do this functionality that you know of? Do they require a keyword argument, or do they automatically detect the type?

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -839,8 +842,9 @@ def _wait_for_msg(self, timeout=0.1):
             pid = pid[0] << 0x08 | pid[1]
             sz -= 0x02
         # read message contents
-        msg = self._sock_exact_recv(sz)
-        self._handle_on_message(self, topic, str(msg, "utf-8"))
+        raw_msg = self._sock_exact_recv(sz)
```
brentru, line 845: Could we automatically detect here somehow, instead?

brentru (changes requested):

adafruit_minimqtt/adafruit_minimqtt.py
alvarowolfx: I'm used to using MQTT with Arduino and things are a bit different because in C, a `char` and a `byte` array are basically the same things. But in CircuitPython, the message is read in a `bytearray`, and I don't know of a way to check beforehand if the content will be a valid Unicode string or if is just a binary blob. In C you could just interpret the `char` array the way you need. 

I don't know of a similar library that does the same. Like I mentioned above, on the Arduino lib that we are using, the default behavior is to use binary messages and the user handles the conversion.

Example: https://github.com/arduino-libraries/ArduinoMqttClient/blob/master/examples/WiFiSimpleReceiveCallback/WiFiSimpleReceiveCallback.ino#L104
This one is even better that allows you to stream the message and read bigger messages that don't fit in RAM.

I can change the flag to `use_binary_mode` here.

adafruit_minimqtt/adafruit_minimqtt.py
alvarowolfx: Not sure how to detect if the `bytearray` content is a Unicode string or a binary blob 😞 . 

We can add a try/except block and convert it to a Unicode string, if it fails it returns the message in binary, but not sure if is consistent behavior because users might have to check the data type on the callback function.

adafruit_minimqtt/adafruit_minimqtt.py
brentru: Thanks for the explanation, let's have the user handle conversion in user-code.

Please change the flag to `use_binary_mode` to make it more descriptive and tag me for a re-review

adafruit_minimqtt/adafruit_minimqtt.py
brentru: Yup, let's have the user handle it in user code instead.

brentru (approved):

PR #99, opened 2022-01-22
Added missing blank lines before :param blocks in docstrings

rimwolf-redux:
In the documentation for adafruit_minimqtt at https://circuitpython.readthedocs.io/projects/minimqtt/en/latest/api.html, the :param information in the class docstring and several method docstrings is run into the description rather than being neatly formatted. This fix only adds blank lines as necessary to enable more readable formatting.


brentru (approved):
Thanks!

PR #106, opened 2022-03-06
add debug logging to recieved messages.

dgriswo:
writes topic and raw message when loop receives a message from a subscribed topic, when debug is enabled.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -237,11 +237,11 @@ class MQTT:
                 "ssl_context must be set before using adafruit_mqtt for secure MQTT."
             )
 
-        if self.logger and port == MQTT_TLS_PORT:
+        if self.logger is not None and port == MQTT_TLS_PORT:
             self.logger.info(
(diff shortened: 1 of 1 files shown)


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -853,6 +853,10 @@ def _wait_for_msg(self, timeout=0.1):
         # read message contents
         raw_msg = self._sock_exact_recv(sz)
         msg = raw_msg if self._use_binary_mode else str(raw_msg, "utf-8")
+        if self.logger:
```
brentru, line 856: Check `if self.logger is not None:` instead, please.

brentru (changes requested):

adafruit_minimqtt/adafruit_minimqtt.py
dgriswo: I updated this and other conditionals to use the is not None:  

I have been able to successfully exercise some of the standard use cases, but none of the error based use cases.

brentru (approved):

PR #116, opened 2022-08-17
allow to set socket timeout

vladak:
This change allows to set socket timeout via the init function, like so:
```python
 mqtt_client = MQTT.MQTT(
        broker=secrets["broker"],
        port=secrets["broker_port"],
        socket_pool=pool,
        socket_timeout=3,
    )
```
I verified with `strace` that the `poll()` syscall timeout value reflects the `socket_timeout` parameter.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -214,12 +214,12 @@ class MQTT:
         self.on_unsubscribe = None
 
     # pylint: disable=too-many-branches
-    def _get_connect_socket(self, host, port, *, timeout):
+    def _get_connect_socket(self, host, port, *, timeout=1):
         """Obtains a new socket and connects to a broker.
 
         :param str host: Desired broker hostname
         :param int port: Desired broker port
-        :param int timeout: Desired socket timeout in seconds
+        :param int timeout: Desired socket timeout, in seconds
         """
         # For reconnections - check if we're using a socket already and close it
         if self._sock:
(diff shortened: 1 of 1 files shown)


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -129,6 +131,7 @@ class MQTT:
     :param socket socket_pool: A pool of socket resources available for the given radio.
     :param ssl_context: SSL context for long-lived SSL connections.
     :param bool use_binary_mode: Messages are passed as bytearray instead of string to callbacks.
+    :param int socket_timeout: socket timeout, in seconds
```
tekktrik, line 134: Would you mind expanding this a little, specifically what this means?  For example, I'm not very familiar with sockets so the description doesn't expand on the variable name enough that I would understand the behavior :)

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -209,12 +214,12 @@ def __init__(
         self.on_unsubscribe = None
 
     # pylint: disable=too-many-branches
-    def _get_connect_socket(self, host, port, *, timeout=1):
+    def _get_connect_socket(self, host, port, *, timeout):
```
tekktrik, line 217: Not really a strong request, but I do like having having keyword arguments have defaults where possible, so that not having to call them out works.  I would added the default value back for that reason.

tekktrik (changes requested):
A couple of suggestions, thanks for the submission!

adafruit_minimqtt/adafruit_minimqtt.py
vladak: Makes sense, added back.

tekktrik:
Thanks!  Looks like that description should get broken up into a couple lines to make the linter happy, but looks ready to go after that!

tekktrik (approved):
Awesome, thanks!

PR #119, opened 2022-08-30
make connect retries configurable

vladak:
This change exposes retry count to MQTT class init.


tekktrik (approved):
Looks good to me!  Thanks!

Issue #120, opened 2022-08-31
Make MQTT.__init__() arguments keyword-only


tekktrik:
All existing usages I could find do, so while this is API-breaking, it shouldn't need much work, and protects against accidental argument mix ups.

PR #121, opened 2022-08-31
Make MQTT__init__ arguments keyword-only

tekktrik:
This is API-breaking, but hopefully makes using the library less prone to errors.  I couldn't find any existing libraries or Learn Guides (save for one that's already broken with regards to API since it hasn't been updated for the last API-breaking change to `MQTT.__init__()`).

Resovles #120

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -141,6 +141,7 @@ class MQTT:
     # pylint: disable=too-many-arguments,too-many-instance-attributes, not-callable, invalid-name, no-member
     def __init__(
         self,
+        *,
         broker,
         port=None,
         username=None,


gamblor21 (approved):
Looks good to me. Just unsure if anything needs to be tweaked to tag it for 8.x since it is technically API breaking?

tekktrik:
I think it's fine, but I'll sit on tonight and see if anyone in #circuitpython-dev knows more about that!

tekktrik:
Converted to draft so it doesn't merge before I can ask about this!

dhalbert:
I agree it's fine, with a major version bump, if you've checked the Learn Guides repo and the library examples.

tekktrik:
Checked the libraries and Learn Guides repositories, we're good!

Issue #108, opened 2022-04-12
loop() doesn't fetch all messages for a topic


calcut:
Still trying to suss out what is going on here.

At the simplest level, I'm seeing that `mqtt_client.loop()` *usually* only fetches a single message from the queue ~~(for a particular topic)~~, even if there are many messages waiting. I assume this is incorrect behaviour

I have seen that sometimes it does a 'catch up' and fetch all of the messages!
Not sure if this is a quirk of AdafruitIO's MQTT broker or this library.

e.g. if I run loop() every 5 seconds, expecting a message from AIO's `time/seconds` topic
```
MQTT time: 1649762379
sleep 5
MQTT time: 1649762380
sleep 5
MQTT time: 1649762381
sleep 5
MQTT time: 1649762382
sleep 5
MQTT time: 1649762383
sleep 5
MQTT time: 1649762384
sleep 5
MQTT time: 1649762385
sleep 5
MQTT time: 1649762386
sleep 5
MQTT time: 1649762387
sleep 5
MQTT time: 1649762388
sleep 5
MQTT time: 1649762389
sleep 5
MQTT time: 1649762390 <--------- Here it is very out of sync with the actual time, but then it suddenly does a catch up
sleep 5
MQTT time: 1649762391
MQTT time: 1649762392
MQTT time: 1649762393
MQTT time: 1649762394
....
MQTT time: 1649762436
MQTT time: 1649762437
MQTT time: 1649762438
MQTT time: 1649762439
sleep 5
MQTT time: 1649762440
sleep 5
MQTT time: 1649762441
sleep 5
```

It looks like if the queue builds up too much, it can cause other issues.
e.g. I'm trying to use MQTT and HTTP requests in the same code, there is an interaction where I see errors like:

```
Traceback (most recent call last):
  File "ssl_debug.py", line 44, in <module>
  File "adafruit_requests.py", line 752, in get
  File "adafruit_requests.py", line 693, in request
  File "adafruit_requests.py", line 546, in _get_socket
RuntimeError: Sending request failed
```
Which seem to be associated with the MQTT queue not being cleared

calcut:
Should have said, this is:
Adafruit CircuitPython 7.2.5 on 2022-04-06; Adafruit Feather ESP32S2 with ESP32S2

and my code to reproduce is:
```
import time
from secrets import secrets
import wifi
import ssl
import socketpool
import adafruit_requests
[...]

Issue #101, opened 2022-02-18
Occasional MemoryError during Adafruit IO loop()


calcut:
Adafruit CircuitPython 7.1.1 on 2022-01-14; Adafruit Feather ESP32S2 with ESP32S2

Circup freeze:
Found device at /Volumes/CIRCUITPY, running CircuitPython 7.1.1.
adafruit_datetime==1.1.6
adafruit_htu31d==1.1.3
adafruit_requests==1.10.6
adafruit_ticks==1.0.1
neopixel==6.2.4
adafruit_bus_device==5.1.4
adafruit_io==5.6.1
adafruit_logging==3.7.5
adafruit_minimqtt==5.2.1
adafruit_register==1.9.8

```
ERROR Traceback (most recent call last):
  File "mcu.py", line 304, in aio_receive
  File "adafruit_io/adafruit_io.py", line 233, in loop
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 805, in loop
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 854, in _wait_for_msg
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 906, in _sock_exact_recv
MemoryError: memory allocation failed, allocating 4294953983 bytes
```

I'm running `io.loop(timeout=0.1)` in mcy.py, line 304
Normally this works fine, but every few minutes I get the above exception.
If I ignore the exception, the code continues to run with no obvious issues, but I wanted to report it anyway.

digging into adafruit_minimqtt.py, and adding a `print(bufsize)` shows me that the error on line 906 occurs when bufsize is less than zero. e.g. -13313. So I can see how the code ended up trying to allocate ~4GB of memory (2^32-13313)

I also notice that the bufsize immediately before the negative one is large and positive e.g. 13365

I could try harder to make a minimum reproducable example if needed.

ladyada:
thats odd, how could you get a request for a negative amount? but yeah, a reproducable example would help a ton!

calcut:
[code.py.txt](https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/files/8105009/code.py
.txt)
Attached is  my minimal example, it seems to be related to:

`io.subscribe_to_time("seconds")`

which means I receive a MQTT message every second.
Not certain its the volume of messages that means it is more likely to show up, or something buggy with those particular messages.

[...]

PR #122, opened 2022-09-05
Fetch all messages in loop, non-blocking mode

calcut:
First PR, so bear with me!

Basically, I've found issues with callling `loop()` too frequently or too infrequently, that I'm trying to address here.
Calling too frequently (with timeout >0) can lead to a nasty MemoryError. https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/issues/101
Calling too infrequently can cause a build up of messages on the broker. https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/issues/108

I've changed the default timeout to 0 (and added error handling of EAGAIN)
This might be a controversial change - but it means the function is really non-blocking as the comments say it should be. (Perhaps there is a history of why it was set to 1 by default?, I'd be happy keeping it at 1 if it prevents issues for others)

I've also made it call `_wait_for_msg()` in a while loop, similar to how the ping() function works.
I've used the 10s `_recv_timeout` to make sure it can't get stuck in the loop.
This means that all waiting messages are received, not just the next in the queue.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -869,7 +869,7 @@ class MQTT:
                 if self.logger is not None:
                     self.logger.debug(
                         f"Loop timed out, message queue not empty after {self._recv_timeout}s"
-                        )
+                    )
                 break
             rcs.append(rc)
 
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -857,10 +857,10 @@ class MQTT:
 
         stamp = time.monotonic()
         self._sock.settimeout(timeout)
-        rcs = []
+        responses = []
         while True:
             rc = self._wait_for_msg()
-            if rc == None: 
+            if rc is None: 
                 break
             if time.monotonic() - stamp > self._recv_timeout:
                 if self.logger is not None:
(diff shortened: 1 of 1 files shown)


tekktrik:
Do you have any code that can reproduce the issues you listed?  I'm going to try to review this and didn't know if you had something like that already.

calcut:
There is code in the 2 issues I linked. The MemoryError may be harder to reproduce (e.g. if it depends on connection latency etc).

I can try to help further if it still isn't clear

FoamyGuy:
I'm not super familiar with the specifics of MQTT so I can't comment much on the details of the change. I did try it out on a device though. 

Using code posted in the issue on a `Feather ESP32-S2 8.0.0beta.1`.

Can confirm I observed the same issue with currently released libraries, mqtt time "falling behind" due to only getting one callback per loop call and request getting out of retries problem. Eventually it "catches up" with a bunch of callbacks at once (in my case it was dozens, I didn't count exactly)

The new version of this library from the PR does seem to resolve it. With this version I am getting fairly consistently 5 mqtt callbacks per loop with the final one matching the time given by the http request. 

I inclined to approve, but we should probably try out a few of the learn guide projects that use minimqtt to ensure the differing functionality with callbacks does not have adverse effects in them.  Assuming no issues with them, I think this is good.

FoamyGuy (approved):
This looks good to me. 

I tested successfully with PyPortal using this learn guide: https://learn.adafruit.com/adafruit-io-a-c-power-relay/circuitpython-setup

And Feather ESP32-S2 TFT with the time subscription reproducer script from the issue. And Fun House using the native networking example inside the examples folder of this repo. 

Everything appears to function as intended to me even with the new behavior that gets all messages in the queue every loop instead of only 1 most loops.

Thanks for working on this fix @calcut

PR #123, opened 2022-09-18
_wait_for_msg(): accept EAGAIN in addition to ETIMEDOUT

rtwfroody:
You get EAGAIN when passing a timeout value of 0 and there is no data in the buffer.


brentru (approved):
lgtm

Issue #126, opened 2022-10-27
Lengthy delays in .loop()


jddmartin:
~~There is a recent commit 48aba3c63fc3384f426bd08ededff7d6c2d4fa98 which causes a lengthy delay in calls to .loop() in one of my test programs, which did not occur prior to this commit.~~

**EDIT**: Apologies, I made a mistake in identifying a specific commit.  The commit tagged 5.5.2 shows the lengthy delay, and commits tagged 5.5.1 and 5.5.0 do not show the problem.

I am using an ethernet featherwing attached to a rp2040 feather.  Running the test code below (a modified version of example programs for minimqtt) before and after commit shows a lengthy (several second) delay introduced in calls to .loop()  (the third last line in code).  Whereas prior to this commit, loop returns quickly; i.e., on order of specified timeout.

I'm running circuitpython 7.3.3, but I get identical slow behaviour with 8.0.0 and most recent version of minimqtt.

```
import board
import busio
from digitalio import DigitalInOut

from adafruit_wiznet5k.adafruit_wiznet5k import WIZNET5K
import adafruit_wiznet5k.adafruit_wiznet5k_socket as socket
import adafruit_minimqtt.adafruit_minimqtt as MQTT

print("starting")

cs = DigitalInOut(board.D10)
spi_bus = busio.SPI(board.SCK, MOSI=board.MOSI, MISO=board.MISO)

print("Finished spi")

# Initialize ethernet interface with DHCP
eth = WIZNET5K(spi_bus, cs)

print("My IP address is:", eth.pretty_ip(eth.ip_address))

print("Finished ethernet")

# pylint: disable=unused-argument
def message(client, feed_id, payload):
    # Message function will be called when a subscribed feed has a new value.
    # The feed_id parameter identifies the feed, and the payload parameter has
    # the new value.
    print("Feed {0} received new value: {1}".format(feed_id, payload))

# Initialize MQTT interface with the ethernet interface
MQTT.set_socket(socket, eth)

# Initialize a new MQTT Client object
mqtt_client = MQTT.MQTT(
    broker="192.168.1.200",
    port = 1883,
    is_ssl = False
#    username=secrets["aio_username"],
#    password=secrets["aio_key"],
)

print("trying to connect")
mqtt_client.connect()

i=0
while True:
    mqtt_client.loop(timeout=0.01)
    i += 1
    print(i)
'''

rpavlik:
[...]

PR #127, opened 2022-11-16
Issue #126

jinglemansweep:
Added bytearray('\x00') to the list of expected "noop" returns from the socket.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -896,7 +896,7 @@ class MQTT:
 
         # Block while we parse the rest of the response
         self._sock.settimeout(timeout)
-        if res in [None, b"", bytearray(b'\x00')]:
+        if res in [None, b"", b"\x00"]:
             # If we get here, it means that there is nothing to be received
             return None
         if res[0] == MQTT_PINGRESP:
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -896,7 +896,7 @@ class MQTT:
 
         # Block while we parse the rest of the response
         self._sock.settimeout(timeout)
-        if res in [None, b""]:
+        if res in [None, b"", bytearray(b'\x00')]:
             # If we get here, it means that there is nothing to be received
             return None
         if res[0] == MQTT_PINGRESP:


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -896,7 +896,7 @@ def _wait_for_msg(self, timeout=0.1):
 
         # Block while we parse the rest of the response
         self._sock.settimeout(timeout)
-        if res in [None, b""]:
+        if res in [None, b"", bytearray(b'\x00')]:
```
dhalbert, line 899: Does it work if this is just `b"\x00"`? Not sure you need to create a bytearray.

dhalbert (changes requested):
THanks! One question.

dhalbert:
Use `pre-commit` or run the code through `black` to get rid of the failure: https://learn.adafruit.com/improve-your-code-with-pylint

adafruit_minimqtt/adafruit_minimqtt.py
jinglemansweep: I've just tried:

    if res in [None, b"", b"\x00"]:

And it appears to work as expected. Decent matrix framerates, no `MemoryError`'s etc.

dhalbert (approved):
I'm wondering whether the fact that you have to check for one byte of zeros is a bug in the underlying network code. But I think this is a good workaround to add to the library for now.

@brentru Do you have any comment about this from an MQTT protocol point of view?

brentru:
@jinglemansweep Hi - thanks for fixing this issue. @dhalbert - I do not think this is a network or a MQTT spec issue, `b"\x00` is rx'd after the call to `res = self._sock_exact_recv(1)`. 

Considering some in #126 are having issues with non-native wifi,  it's possible (don't know, guessing right now) that the issue is within this library's local implementation of `_sock_exact_recv()` vs the native impl. of `_sock_exact_recv()` (cc @calcut ☝️ )

```
        # CPython socket module contains a timeout attribute
        if hasattr(self._socket_pool, "timeout"):
            try:
                res = self._sock_exact_recv(1)
            except self._socket_pool.timeout:
                return None
        else:  # socketpool, esp32spi
            try:
                res = self._sock_exact_recv(1)
```

brentru (approved):
For now, though, I think this is an OK temporary change and will cause clients to poll faster until we resolve why the socket is rcv'ing a bytearray on some clients.

Issue #131, opened 2022-12-01
Memory allocation error on Raspberry Pi Pico W?


ilikecake:
I apologize if this is just an error on my part, but I am having trouble getting this library to work with a Raspberry Pi Pico W. Please see [this thread](https://forums.adafruit.com/viewtopic.php?t=196487) for details. It appears that code is attempting to allocate a negative memory size, which is obviously a problem. I don't know enough about the MQTT protocol or debugging Wifi to find the issue.

Is this a bug in the MQTT library, or is my code the problem?

vladak:
This is curious. It seems that the piece of code that is annotated with variable values in the forum in `_wait_for_msg()` should only be executed for the `PUBLISH` message type, i.e. when the server publishes a message to the client. Here, it seems that it is trying to interpret variable payload that is not consistent with the message type. The length 12320 corresponds to bytes 0x30 0x20. Anyway, it is wildly improbable that the server uses such a long topic length.

For `connect()`, it should return from `_wait_for_msg()` right here: https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/4c0c7706b1aab9301c608ba0241953702b280ee1/adafruit_minimqtt/adafruit_minimqtt.py#L912-L913 i.e. before the code that lead the exception.

Note that the line numbers do not match as I am using the latest code, but you get the idea.

It would make sense to get the value of `res[0]` and ideally also complete traffic dump for the TCP connection between the MQTT client and the broker when this problem happens. Also, I wonder what kind of software is the MQTT broker running.

Anyhow, the code should be more resilient against internally inconsistent messages.

vladak:
There is some sort of problem when a MQTT message spans multiple TCP segments. Only the data from the first segment is read in a single `_wait_for_msg()` call, the rest of the buffer is filled with zero bytes. The remaining segments of the message payload are read in the next `_wait_for_msg()` call and are interpreted as MQTT headers, causing all sorts of mayhem.

When connected to `test.mosquitto.org` and subscribed to the `#` topic, adafruit_minimqtt does not survive more than couple of seconds.

vladak:
[...]

PR #132, opened 2022-12-03
make _wait_for_msg() more readable

vladak:
Inspired by #131. Will not fix the actual problem.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -47,6 +47,7 @@ MQTT_TLS_PORT = const(8883)
 # MQTT Commands
 MQTT_PINGREQ = b"\xc0\0"
 MQTT_PINGRESP = const(0xD0)
+MQTT_PUBLISH = const(0x30)
 MQTT_SUB = b"\x82"
 MQTT_UNSUB = b"\xA2"
 MQTT_DISCONNECT = b"\xe0\0"
@@ -879,7 +880,9 @@ class MQTT:
     def _wait_for_msg(self, timeout=0.1):
         # pylint: disable = too-many-return-statements
 
-        """Reads and processes network events."""
+        """Reads and processes network events.
+        Return the packet type or None if there is nothing to be received.
+        """
         # CPython socket module contains a timeout attribute
         if hasattr(self._socket_pool, "timeout"):
             try:
@@ -909,8 +912,11 @@ class MQTT:
                     "Unexpected PINGRESP returned from broker: {}.".format(sz)
                 )
             return MQTT_PINGRESP
-        if res[0] & 0xF0 != 0x30:
+
+        if res[0] & 0xF0 != MQTT_PUBLISH:
             return res[0]
+
+        # Handle only the PUBLISH packet type from now on.
         sz = self._recv_len()
         # topic length MSB & LSB
         topic_len = self._sock_exact_recv(2)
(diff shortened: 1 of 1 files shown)


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -909,12 +911,21 @@ def _wait_for_msg(self, timeout=0.1):
                     "Unexpected PINGRESP returned from broker: {}.".format(sz)
                 )
             return MQTT_PINGRESP
-        if res[0] & 0xF0 != 0x30:
+
+        if res[0] & 0xF0 != MQTT_PUBLISH:
```
brentru, line 915: Since it was made a `const()`, could you check if we use `0x30` elsewhere in the file and replace with `MQTT_PUBLISH` ?

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -909,12 +911,21 @@ def _wait_for_msg(self, timeout=0.1):
...
             return res[0]
+
+        # Handle only the PUBLISH packet type from now on.
         sz = self._recv_len()
         # topic length MSB & LSB
         topic_len = self._sock_exact_recv(2)
         topic_len = (topic_len[0] << 8) | topic_len[1]
+
+        if topic_len > sz - 2:
```
brentru, line 924: What's the rationale for `sz - 2`?

brentru (changes requested):

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -900,7 +902,7 @@ def _wait_for_msg(self, timeout=0.1):
         if res in [None, b"", b"\x00"]:
             # If we get here, it means that there is nothing to be received
             return None
-        if res[0] == MQTT_PINGRESP:
+        if res[0] & 0xF0 == MQTT_PINGRESP:
```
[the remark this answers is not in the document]
vladak: It's masking out the reserved part of the first byte in the fixed MQTT header. Here it is used to get the MQTT command type. Will `const()` it.

adafruit_minimqtt/adafruit_minimqtt.py
vladak: Sure, done.

adafruit_minimqtt/adafruit_minimqtt.py
vladak: The `sz` is the "remaining length" from the MQTT fixed (sic!) header. It specifies how many bytes will follow. For the PUBLISH type, a variable header follows the fixed header. The variable header should always start with a topic name. The topic name should be [encoded as UTF-8 string](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_UTF-8_Encoded_String) which means the length will be represented as 2 bytes (MSB and LSB - as also visible in the code), unlike the variable encoding used in the fixed header for the remaining length (hence the sic).

If the 2 bytes say that the topic length is bigger than the remaining length sans the 2 bytes, the packet is internally inconsistent.

For a MQTT packet with empty message, where remaining length is R, the topic length should be equal R - 2.

adafruit_minimqtt/adafruit_minimqtt.py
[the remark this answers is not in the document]
brentru: Thanks for clarifying

brentru (approved):
LGTM!

PR #133, opened 2022-12-05
make sure to read the whole buffer in _sock_exact_recv()

vladak:
This change addresses the problem of returning incomplete buffer from _sock_exact_recv() in CPython.

Tested with:
```python
#!/usr/bin/env python3

import socket
import ssl
import logging


def connect_hook(client, user_data, result, code):
    print(f"Connect: {user_data} {result} {code}")


def message_hook(client, topic, message):
    print(f"Message: topic='{topic}' message='{message}'")


broker = "test.mosquitto.org"
port = 1883

mqtt_client = MQTT.MQTT(
    broker=broker,
    port=port,
    socket_pool=socket,
)


logging.basicConfig()
# mqtt_client.enable_logger(logging, log_level=logging.DEBUG)

# mqtt_client.on_connect = connect_hook
mqtt_client.on_message = message_hook

mqtt_client.connect()
mqtt_client.subscribe('#')

while True:
    try:
        mqtt_client.loop(5)
    except UnicodeDecodeError:
        pass
```

It survived the wild message traffic there for bunch of minutes.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -965,7 +965,11 @@ class MQTT:
             mv = memoryview(rc)
             recv_len = self._sock.recv_into(rc, bufsize)
             to_read = bufsize - recv_len
-            assert to_read >= 0
+            if to_read < 0:
+                raise MMQTTException(
+                    f"negative number of bytes to read: "
+                    f"{to_read} = {bufsize} - {recv_len}"
+                )
             read_timeout = self.keep_alive
             mv = mv[recv_len:]
             while to_read > 0:
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -963,15 +963,15 @@ class MQTT:
             # CPython/Socketpool Impl.
             rc = bytearray(bufsize)
             mv = memoryview(rc)
-            recv = self._sock.recv_into(rc, bufsize)
-            to_read = bufsize - recv
+            recv_len = self._sock.recv_into(rc, bufsize)
+            to_read = bufsize - recv_len
             assert to_read >= 0
             read_timeout = self.keep_alive
-            mv = mv[recv:]
+            mv = mv[recv_len:]
             while to_read > 0:
-                recv = self._sock.recv_into(mv, to_read)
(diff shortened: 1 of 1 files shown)


dhalbert:
> Eventually it got `OSError` (`ETIMEDOUT`) however that is probably not related to the changes.

What do you think this is related to? Do you think the server is doing rate limiting or something?

vladak:
> > Eventually it got `OSError` (`ETIMEDOUT`) however that is probably not related to the changes.
> 
> What do you think this is related to? Do you think the server is doing rate limiting or something?

When I run this test with CPython on my laptop connected WiFi hotspot on my phone that is using 4G to connect to the Internet, it just continuously runs until interrupted. Given that when the (slightly adapted) test code runs on the QtPy, using WiFi AP that is connected to super stable Ethernet based Internet upstream and considering it uses the same branch in ` _sock_exact_recv()`, I'd attribute this to WiFi flakiness / network delays and/or to the lower layers (CircuitPython / the OS running on the chip). The way I look at this change is that this fixes a problem that allows to uncover more problems underneath.

dhalbert:
@brentru Could you take a look at this?

brentru:
@dhalbert  is there urgency? I am out today, returning tomorrow.

dhalbert:
> @dhalbert is there urgency? I am out today, returning tomorrow.

NP,  this was just hanging for a while.

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -985,15 +956,27 @@ def _sock_exact_recv(self, bufsize):
...
         if not self._backwards_compatible_sock:
             # CPython/Socketpool Impl.
             rc = bytearray(bufsize)
-            self._sock.recv_into(rc, bufsize)
-        else:  # ESP32SPI Impl.
-            stamp = time.monotonic()
+            mv = memoryview(rc)
+            recv_len = self._sock.recv_into(rc, bufsize)
+            to_read = bufsize - recv_len
+            assert to_read >= 0
```
brentru, line 968: Want to raise a descriptive exception here, instead?

vladak: Sure, it will provide more data that can be used for debugging.

brentru (approved):
Thanks for making the change I requested! LGTM

PR #137, opened 2022-12-26
add type hints to matcher.py

vladak:
This change approaches #92, specifically for `matcher.py`. After that `mypy` complains only about the function parameters/assignment:
```
$ mypy --strict adafruit_minimqtt/matcher.py 
adafruit_minimqtt/matcher.py:50: error: Incompatible types in assignment (expression has type "Callable[..., Any]", variable has type "None")  [assignment]
adafruit_minimqtt/matcher.py:100: error: Incompatible types in assignment (expression has type "None", variable has type "Callable[..., Any]")  [assignment]
Found 2 errors in 1 file (checked 1 source file)
```

This is my first time adding type hints to Python code and have to say, being used to strongly typed language, that it feels a bit strange - retrofitting type specification that is still limping, e.g. the case of `None` or missing function type parameters. That said, here you go.

### DIFF
--- a/adafruit_minimqtt/matcher.py
+++ b/adafruit_minimqtt/matcher.py
@@ -11,6 +11,13 @@ https://github.com/eclipse/paho.mqtt.python/blob/master/src/paho/mqtt/matcher.py
 * Author(s): Yoch (https://github.com/yoch)
 """
 
+try:
+    from typing import Dict, Any
+except ImportError:
+    pass
+
+from collections.abc import Callable, Iterator
+
 
 class MQTTMatcher:
     """Intended to manage topic filters including wildcards.
@@ -27,14 +34,14 @@ class MQTTMatcher:
 
         __slots__ = "children", "content"
 
-        def __init__(self):
-            self.children = {}
+        def __init__(self) -> None:
+            self.children: Dict[str, MQTTMatcher.Node] = {}
             self.content = None
 
-    def __init__(self):
+    def __init__(self) -> None:
         self._root = self.Node()
 
-    def __setitem__(self, key, value):
+    def __setitem__(self, key: str, value: Callable[..., Any]) -> None:
         """Add a topic filter :key to the prefix tree
         and associate it to :value"""
         node = self._root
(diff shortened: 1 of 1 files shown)


adafruit_minimqtt/matcher.py
```diff
@@ -11,6 +11,13 @@
 * Author(s): Yoch (https://github.com/yoch)
 """
 
+try:
+    from typing import Any, Dict
+except ImportError:
+    pass
+
+from collections.abc import Callable, Iterator
```
tekktrik, line 19: CircuitPython doesn't support `collections.abc`, so this should go after the `typing` import.  Additionally, it looks like there was a bug with using `collections.abc.Callable` for typing in 3.9.1, so I would just use `typing.Callable` and `typing.Iterator`.

adafruit_minimqtt/matcher.py
```diff
@@ -27,22 +34,22 @@ class Node:
...
+        def __init__(self) -> None:
+            self.children: Dict[str, MQTTMatcher.Node] = {}
             self.content = None
 
-    def __init__(self):
+    def __init__(self) -> None:
         self._root = self.Node()
 
-    def __setitem__(self, key, value):
+    def __setitem__(self, key: str, value: Callable[..., Any]) -> None:
```
tekktrik, line 44: Since it's `Callable[..., Any]`, it can be shortened to just `Callable`.

tekktrik (changes requested):
Thanks for the submission!  A couple notes of feedback:

adafruit_minimqtt/matcher.py
vladak: okay, should be fixed.

adafruit_minimqtt/matcher.py
vladak: `mypy` does not like bare Callable:
```
Missing type parameters for generic type "Callable"
```
Also, I was thinking that with the future type hint additions to the main module the types of the arguments can be back filled.

tekktrik: On second thought, let's leave that unannotated for now so it doesn't look mistakenly annotated.  it may even be a good idea to leave a commented to-do so that when annotations are finished this should be revisited.

vladak: okay, Callable removed. Will add a note to the issue tracking the type hints.

tekktrik (approved):
Awesome, thanks!

PR #139, opened 2023-01-08
do not hard-code TLS port value

vladak:
When experimenting with TLS based connections to various ports on https://test.mosquitto.org/ , I found that the 8883 value is hard-coded. This change replaces the values with the global const. This does not fix the issue with specifying arbitrary port as TLS port, it merely straightens the use of global constant.


brentru (approved):
This looks good to me, utilization of const() is always appreciated!

PR #141, opened 2023-01-08
use correct terminology for enable_logger()

vladak:
I noticed that the documentation for `enable_logger()` talks about logger object, while it actually means logging package to create/retrieve the logger. While there I added the possibility to change logger name because by default such generic name was used so it might conflict with a logger created before interacting with `adafruit_minimqtt` (also given how logger creation in `adafruit_logging` works).

While there, I made the function return the logger object, in case someone wants to further change it (e.g. add handlers). I did this because `self.logger` is not preceded with `_` so probably not considered private.


brentru (approved):
Thank you for the PR!

PR #143, opened 2023-01-12
remove unused doc reference to network manager

vladak:
There is no such argument as `network_manager` in `MQTT()` so this change updates the doc string.


tekktrik (approved):

tekktrik:
Thanks!

PR #150, opened 2023-02-05
use NullHandler for logging by default

vladak:
This change makes sure the `self.logger` is always set to something. Given that `enable_logger()` has the flexibility of specifying the logging package, the loggers are switched, as opposed to handler juggling which might not be a good idea anyway as multiple log packages can be used simultaneously.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,7 @@ Dependencies
 This driver depends on:
 
 * `Adafruit CircuitPython <https://github.com/adafruit/circuitpython>`_
+* `Adafruit CircuitPython Logging <https://github.com/adafruit/Adafruit_CircuitPython_Logging>`_
 
 Please ensure all dependencies are available on the CircuitPython filesystem.
 This is easily achieved by downloading
--- a/README.rst
+++ b/README.rst
@@ -24,7 +24,6 @@ Dependencies
 This driver depends on:
 
 * `Adafruit CircuitPython <https://github.com/adafruit/circuitpython>`_
-* `Adafruit CircuitPython Logging <https://github.com/adafruit/Adafruit_CircuitPython_Logging>`_
 
 Please ensure all dependencies are available on the CircuitPython filesystem.
 This is easily achieved by downloading
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -202,7 +202,7 @@ class MQTT:
         self._pid = 0
         self._timestamp = 0
         self.logger = NullLogger()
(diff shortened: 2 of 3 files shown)


tekktrik (commented):
This looks good to me, and seems functionally the same.  My only concern would be memory issues (e.g., `adafruit_logging` takes up precious RAM needed by a board).  It may be worth seeing if this works on an M0 board.  @brentru, was that a concern you had when making the library?

vladak:
That's interesting and I have not considered that. Is there perhaps some tooling that would help with memory analysis ?

FoamyGuy:
Can you expand a bit on why the way it's done today with the None check is not a good idea? I don't know what you meant by handler juggling. 

This does seem to me like it would be functionally the same, but at the "cost" of extra RAM and the additional dependency needing to be loaded. 

I don't quite understand why having the NullHandler pre-set by default like this preferable to having it be None and using the if statements to determine whether it should try to log or not.

vladak:
It's a readability issue and I'd hope that the code changes make it apparent. Currently in the code there are 31 checks that determine whether `self.logger` is set to `None` so that the logger can be actually used. It is also a stability issue - when I was doing some changes to the code recently, I had to be carefull to carry the `logger is not None` checks around and also make sure that any new logger use is underneath the check. If I was not careful with keeping the check for all the cases where logger is used, then it might not be caught in my cursory testing that had the logger always set to something. It's similar (pattern wise) to the `isLoggable()` check done in Java logging to avoid unnecessary CPU churn if the log level in effect does not match the log level of the log statement, however in the case of the `None` check it results in NoneType exception rather than slower code.

As for the handler juggling, it was alternative idea when I considered the possibilities of addressing the readability issue. Basically have a single logger with NullHandler and replace it with handlers from the logger passed to `enable_logger()` and reverse the operation in `disable_logger()`. This approach also needs a logging module imported, obviously, is more complicated than necessary and does not play well with multiple distinct logging packages (say adafruit logging with CPython logging).

Another possibility would be to wrap this inside a function.

dhalbert (approved):
This looks good to me. @tekktrik @FoamyGuy What do you think?

tekktrik:
I think this is good as well.  It's then probably worth adding a docstring to the `logger` attribute in a follow-up PR the logging option is documented.

dhalbert:
> I think this is good as well. It's then probably worth adding a docstring to the `logger` attribute in a follow-up PR the logging option is documented.

Not merged yet so could be added in this PR.

vladak:
> @vladak is that something you mind doing? 

No issue other than my sleep schedule :-)

tekktrik (approved):
Thanks!

PR #151, opened 2023-02-07
exponential backoff for (re)connect

vladak:
This change implements exponential back-off for connect/reconnect based on the Google IoT algorithm description (sic!), with a nuance in handling temporary errors (like name resolution errors or memory shortage). At one point I had code that applied the exponential back-off also for successfull reconnect()s, however found it a bit too complex and hand-holdy. If one calls reconnect() in a loop even after prior successful connect without sleeping in between, he probably deserves to be rate limited by the broker.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -77,6 +77,10 @@ class MMQTTException(Exception):
     # pass
 
 
+class TemporaryError(Exception):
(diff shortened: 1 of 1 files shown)


FoamyGuy (approved):
This looks good to me to. Thank you @vladak for implementing the backoff logic. 

I tested the changes successfully with a Feather ESP32-S2 TFT using the examples in this repo and adafruit.io as the server side.

PR #152, opened 2023-02-14
add missing logger check

vladak:
As mentioned in https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/pull/150#issuecomment-1429391988 the logger check is error prone.


dhalbert (approved):
Thanks! Good for now, looking forward to #150.

PR #155, opened 2023-02-16
_recv_into() is not used anymore

vladak:
This change removes `_recv_into()`. It was initially introduced in 8492a58f88465148268546fd58712c41208dee6c, then actually used in 2ae61a459dd2842981a15b63f0e60c0a84646749 and then the call was removed in ea966339379c8081dba94c95bd1f286dfe4dcca1.


FoamyGuy (approved):
Looks good to me. Thank you @vladak.

I checked 'find usages' in pycharm to confirm this was unused and indeed it comes up empty. I also verified aio example script still working as intended with Feather ESP32-S2 TFT

PR #156, opened 2023-02-16
handle variable payload in SUBACK

vladak:
As explained in https://github.com/adafruit/circuitpython/issues/7606, the SUBACK/UNSUBACK deserves more robust handling.


FoamyGuy (approved):
Looks good to me. 

I did read through the associated core thread, the specifics within the mqtt protocol are outside my experience. 

I was able to test this version successfully with Adafruit.io on a Feather ESP32-S2 TFT subscribing and sending/receiving messages is working as intended.

PR #162, opened 2023-04-25
Settings dot toml

BlitzCityDIY:
hihi @brentru - I updated the examples that are in the MiniMQTT guide to use settings.toml. Had to change up the WiFi connection method for ESP32SPI since adafruit_esp32spi_wifimanager assumes secrets is being used. Tested native networking with a QT Py ESP32-S2 and ESP32SPI with a Metro M7.

if you are good with how these are formatted I can update the other examples as well.

### DIFF
--- a/examples/minimqtt_simpletest.py
+++ b/examples/minimqtt_simpletest.py
@@ -31,7 +31,7 @@ mqtt_topic = "test/topic"
 
 # Adafruit IO-style Topic
 # Use this topic if you'd like to connect to io.adafruit.com
-# mqtt_topic = secrets["aio_username"] + '/feeds/temperature'
+# mqtt_topic = aio_username + '/feeds/temperature'
 
 
 ### Code ###
@@ -77,8 +77,8 @@ pool = socketpool.SocketPool(wifi.radio)
 mqtt_client = MQTT.MQTT(
     broker="io.adafruit.com",
     port=8883,
-    username=os.getenv("CIRCUITPY_WIFI_SSID"),
-    password=os.getenv("CIRCUITPY_WIFI_PASSWORD"),
+    username=aio_username,
+    password=aio_key,
     socket_pool=pool,
     ssl_context=ssl.create_default_context(),
 )
--- a/examples/minimqtt_simpletest.py
+++ b/examples/minimqtt_simpletest.py
@@ -2,39 +2,57 @@
 # SPDX-License-Identifier: MIT
 
 import os
-import ssl
-import socketpool
-import wifi
+import board
+import busio
+from digitalio import DigitalInOut
+from adafruit_esp32spi import adafruit_esp32spi
+import adafruit_esp32spi.adafruit_esp32spi_socket as socket
 import adafruit_minimqtt.adafruit_minimqtt as MQTT
 
 # Add settings.toml to your filesystem CIRCUITPY_WIFI_SSID and CIRCUITPY_WIFI_PASSWORD keys
-# with your WiFi credentials. DO NOT share that file or commit it into Git or other
-# source control.
+# with your WiFi credentials. Add your Adafruit IO username and key as well.
+# DO NOT share that file or commit it into Git or other source control.
 
-# Set your Adafruit IO Username and Key in settings.toml
-# (visit io.adafruit.com if you need to create an account,
-# or if you need your Adafruit IO key.)
 aio_username = os.getenv("aio_username")
 aio_key = os.getenv("aio_key")
 
-print("Connecting to %s" % os.getenv("CIRCUITPY_WIFI_SSID"))
-wifi.radio.connect(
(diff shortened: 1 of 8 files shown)


examples/minimqtt_simpletest.py
```diff
@@ -1,30 +1,29 @@
...
-aio_username = secrets["aio_username"]
-aio_key = secrets["aio_key"]
-
-print("Connecting to %s" % secrets["ssid"])
-wifi.radio.connect(secrets["ssid"], secrets["password"])
-print("Connected to %s!" % secrets["ssid"])
+aio_username = os.getenv("aio_username")
+aio_key = os.getenv("aio_key")
+aio_broker = os.getenv("broker")
+aio_port = os.getenv("port")
```
brentru, line 20: @BlitzCityDIY  Do we specify in any guides that `aio_port` and `aio_broker` are required in the TOML file? The comments on L14 specifies only IO username and password are set within this file.

I'm open to defining the MQTT client like below to make it more transparent to the user of our MQTT client that a) it's not IO-specific and b) where exactly this sketch is connecting to:

```
mqtt_client = MQTT.MQTT(
    broker="io.adafruit.com",
    port=8883,
    username=os.getenv("CIRCUITPY_WIFI_SSID"),
    password=os.getenv("CIRCUITPY_WIFI_PASSWORD"),
    socket_pool=pool,
)
```

BlitzCityDIY: I agree, I think that makes sense. I was mirroring the secrets syntax but having it be defined like that will be easier for folks to understand I think.

examples/minimqtt_simpletest.py
```diff
@@ -77,10 +77,10 @@ def message(client, topic, message):
 
 # Set up a MiniMQTT Client
 mqtt_client = MQTT.MQTT(
-    broker=secrets["broker"],
-    port=secrets["port"],
-    username=secrets["aio_username"],
-    password=secrets["aio_key"],
+    broker=aio_broker,
+    port=aio_port,
+    username=os.getenv("CIRCUITPY_WIFI_SSID"),
```
brentru, line 80: You have the WiFi SSID and pass here, these should be the IO keys.

brentru:
@BlitzCityDIY  When this has been updated and is ready. Under "Reviewers" please click the ♻️  next to my username that says "re-request review"

BlitzCityDIY:
hihi @brentru - i went through and retested this PR and made some tweaks. all native networking examples were tested with a feather esp32-s2 and the esp32spi examples were tested with a metro m7. i'm at a loss though on the two black errors. i've run pre-commit and black locally with both windows command line and WSL and everything is passing but CI keeps flagging those two examples for black errors.

brentru:
@tekktrik @FoamyGuy Not sure if you are the correct people to ping here. @BlitzCityDIY needs some help with the Black issues blocking this PR from being merged. Could you please take a look into this?

brentru (approved):
Code looks good. Updated to use the settings.toml and using `f` strings for the print statements is a welcome touch!

tekktrik:
I'm unavailable this weekend but can take a look afterwards!

FoamyGuy:
I think I ran across this recently somewhere else too. I figured it was something weird about my local environment, but perhaps it's a more widespread issue.

It seems that for some reason the Black formatting can have different results when it's running inside the actions container vs. when it runs locally.

This PR branch is passing pre-commit for me locally:
image

But failed in the actions check:
```
Error: black....................................................................Failed
- hook id: black
- files were modified by this hook

Error: reformatted examples/esp32spi/minimqtt_pub_sub_nonblocking_esp32spi.py
Error: reformatted examples/native_networking/minimqtt_pub_sub_blocking_native_networking.py

All done! ✨ 🍰 ✨
2 files reformatted, 21 files left unchanged.
```

It looks like the version specified in the pre-commit config was changed recently-ish (~2 months) but this PR branch perhaps was made before that, so it's config file is still on the old one. Then I guess when it runs in actions it's running under the "wrong" one, or a different one from main or something.

@tekktrik I had a look through the actions log and didn't find anywhere that it printed the version of black that it actually used. It printed versions for many other things, but I think not black. If possible / easy it might be nice if we can add `black --version` into the actions task somewhere so that it'll pipe that through into the logs to make it easier to figure out when anything like this difference result in local vs. remote that pops up.

FoamyGuy:
It was indeed a difference of versions in black. It seems that when it runs in the actions container it implicitly merges main before the checks run. But when it's run locally there is nothing enforcing that behavior.

After I merged main I was able to get the same results as the actions check. The latest commits bring in main, and the changes that black formatter wanted. 

Should be good to go if you want to merge this one now @brentru.

Thanks for working on this @BlitzCityDIY!

tekktrik:
I can add a step in the build workflow to read the `.pre-commit-config.yaml` file and print out relevant versions being used!

PR #172, opened 2023-07-24
add basic loop() test

vladak:
Since finishing PR #169, I have been thinking about at least simple test for the loop() return semantics. Here it is.

There is some room for improvement (like always). Specifically, there are other aspects of the function that could be tested (such as MQTT ping) and also if loop() was not cooperating (by returning within some time), the test would be hosed.


tekktrik (approved):
Test code is best code!  And more tests are always good, if you do end up making more.

PR #174, opened 2023-08-25
check is_connected in loop()

vladak:
[Someone hit a None-deref](https://forums.adafruit.com/viewtopic.php?t=203980) when calling `loop()` without calling `connect()` first. This change checks the connected status in loop() and raises exception if not connected.


FoamyGuy (approved):
Makes sense, and looks good to me. Thanks for tracking it down and adding the check @vladak!

I did a quick test of an example script on a Fun House and it's working as expected. I also confirmed calling `loop()` before `connect()` results in the exception being raised.

PR #182, opened 2023-10-30
use time.monotonic_ns() when available

vladak:
This change will use `time.monotonic_ns()` if available to provide monotonic time based counter with consistent precision in order to be able to work for longer periods of time. On systems that lack this API the `__init__()` init will throw `MMQTTException` unless the `use_imprecise_time` is set to True to avoid unpleasant surprises when running for days on end.

At this stage this is being offered for discussion. Once an approach is found suitable, I will perform testing on HW.


vladak:
Casually tested on `Adafruit CircuitPython 8.2.6 on 2023-09-12; Adafruit QT Py ESP32-S3 no psram with ESP32S3` by calling `loop(1)` in an infinite cycle after subscribing to a wildcard topic. Set the logger to one with debug level to observe PING messages etc.

PR #183, opened 2023-11-06
_wait_for_msg() should return message type

vladak:
While traversing the `_wait_for_msg()` code, I noticed that there is discrepancy between what the docs say about the return value and reality. This change fixes that.


FoamyGuy (approved):
Makes sense to me, it was a little akward that sometimes it was returning masked values and other times un-masked. The changes here make it consistent.


This looks good to me. I tested this version successfully on a FunHouse Adafruit CircuitPython 9.0.0-alpha.5

PR #188, opened 2023-11-24
make user_data "public"

vladak:
This change makes using `user_data` more accessible for received message callbacks.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -419,6 +419,8 @@ class MQTT:
 
         Expected method signature is ``on_message(client, topic, message)``
         To get access to the user_data, use the client argument.
+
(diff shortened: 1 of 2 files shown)


FoamyGuy (approved):
This looks good to me. Thank you @vladak. 

I do agree with the overall premise that it would be ideal if the  `on_message()` callback matched the signature of all the other callbacks with regards to getting user data passed to it. It stands out as a discrepancy amongst the different callbacks. 

Honestly I think in the long term I'd be okay with a breaking change that added it as as the second argument so that the callbacks would match signatures from then on. If you're interested in potentially moving that direction maybe it's worth dropping a note in the meeting doc to discuss it in the weeds to get thoughts from the wider team and community.

I tested this version successfully on a FunHouse device with a slightly modified simpletest using some of the code in the new user_data example.

PR #191, opened 2023-11-29
Create minimqtt_multipub_simpletest.py

DJDevon3:
Adds a simple demo for publishing multiple feeds. Designed to be used with the BME280 sensor but filled in with fake values so it should work as is. Most other demos are using the ESP32SPI library.  This one is more designed for the newer ESP32-S2 & S3 with no coprocessor.

### DIFF
--- a/examples/minimqtt_multipub_simpletest.py
+++ b/examples/minimqtt_multipub_simpletest.py
@@ -1,0 +1,139 @@
+# SPDX-FileCopyrightText: 2023 DJDevon3
+# SPDX-License-Identifier: MIT
+# MQTT Multi-Feed Publish Example
+# Coded for Circuit Python 8.2.x
+
+import traceback
+import os
+import time
+import board
+import ssl
+import wifi
+import socketpool
+import adafruit_minimqtt.adafruit_minimqtt as MQTT
+from adafruit_minimqtt.adafruit_minimqtt import MMQTTException
+# from adafruit_bme280 import basic as adafruit_bme280
+
+# Initialize Web Sockets (This should always be near the top of a script!)
+# There can be only one pool
+pool = socketpool.SocketPool(wifi.radio)
+
+# Add settings.toml to your filesystem CIRCUITPY_WIFI_SSID and CIRCUITPY_WIFI_PASSWORD keys
+# with your WiFi credentials. Add your Adafruit IO username and key as well.
+# DO NOT share that file or commit it into Git or other source control.
+ssid = os.getenv("CIRCUITPY_WIFI_SSID")
+appw = os.getenv("CIRCUITPY_WIFI_PASSWORD")
+aio_username = os.getenv("aio_username")
+aio_key = os.getenv("aio_key")
+
+# MQTT Topic
+# Use this format for a standard MQTT broker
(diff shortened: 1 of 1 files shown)


examples/minimqtt_multipub_simpletest.py
```diff
@@ -0,0 +1,130 @@
...
+            time.sleep(10)
+    print("| ✅ WiFi!")
+
+    while wifi.radio.ipv4_address:
+        try:
+            # Connect to MQTT Broker
+            mqtt_client.connect()
+            mqtt_client.publish(feed_01, BME280_temperature)
+            # slight delay required between publishes!
+            # otherwise only the 1st publish will succeed
```
vladak, line 112: That's curious. Do you have any idea why ?

examples/minimqtt_multipub_simpletest.py
```diff
@@ -0,0 +1,130 @@
...
+        except ConnectionError as e:
+            print("Connection Error:", e)
+            print("Retrying in 10 seconds")
+            time.sleep(10)
+    print("| ✅ WiFi!")
+
+    while wifi.radio.ipv4_address:
+        try:
+            # Connect to MQTT Broker
+            mqtt_client.connect()
```
vladak, line 109: This is going to connect (and later) disconnect from the broker repeatedly. This is most likely undesired - single TCP connection is better in terms of resource utilization. The right approach is to connect first and then enter a cycle with `loop()` inside.

examples/minimqtt_multipub_simpletest.py
DJDevon3: No, I'm unsure why that happens.  I only found the workaround by giving it a delay.  Was quite puzzled about that myself.

examples/minimqtt_multipub_simpletest.py
DJDevon3: Capture

I figured that was the right approach as AdafruitIO /overview seems to like that approach.

Also in this screenshot if the delay isn't provided then only the first publish is recorded.  I'll assume AdafruitIO wants them somewhat spaced apart with delays.

vladak: Calling loop() is necessary to avoid being disconnected from the server once the keep alive timer expires.

DJDevon3: This is a bit confusing as that's not how the [simpletest example](https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/main/examples/minimqtt_simpletest.py) works. Connect, publish, disconnect works fine even without increasing the keep alive timer manually in mqtt_client setup.  loop is mainly for checking for messages which I'm not at all interested in if all I want to do is connect & publish. If that's not how it should be done then the simpletest example needs to be corrected to reflect that and the [RTD for MQTT.loop()](https://docs.circuitpython.org/projects/minimqtt/en/stable/api.html#adafruit_minimqtt.adafruit_minimqtt.MQTT.loop) needs to be updated with that bit of information.

I believe it's best practice to disconnect explicitly because I've run into an issue where keeping that connection open while using it with multiple other API's can end up running out of sockets.  Disconnecting closes the socket to be reused by another API.

PR #197, opened 2023-12-22
Switch to using ConnectionManager

justmobilize:
Update `adafruit_requests` to use [Adafruit_CircuitPython_ConnectionManager](https://github.com/adafruit/Adafruit_CircuitPython_ConnectionManager/) for socket handling. As well as clean up examples


requirements.txt
```diff
@@ -3,3 +3,4 @@
 # SPDX-License-Identifier: Unlicense
 
 Adafruit-Blinka
+Adafruit-Circuitpython-ConnectionManager@git+https://github.com/justmobilize/Adafruit_CircuitPython_ConnectionManager@connection-manager
```
vladak, line 6: This sort of stuff makes me wary as it might signal that this library might not receive enough maintenance in the long term. Having something published as a official package would be much more trustworthy in my opinion.

justmobilize: This is a work in progress PR, it will use `Adafruit-Circuitpython-ConnectionManager` when it is released.

justmobilize: Since `ConnectionManager` is merged. This has been changed to pull from pypi

conftest.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: 2023 Justin Myers for Adafruit Industries
```
brentru, line 1: What is the point of this file? Is it run within any workflow? Locally on a device? do you have a future plan for it?

brentru (changes requested):
@justmobilize  This looks OK - did you test any of the tests on physical hardware, or on CPython?

conftest.py
justmobilize: conftest is for tests to build fixtures to reduce boiler plate. I have more in the `request` one here: https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/main/tests/conftest.py. I realize I moved that on into the test directory, and will do the same here

justmobilize:
> @justmobilize This looks OK - did you test any of the tests on physical hardware, or on CPython?

@brentru I've tested on my `m4-express` with both a `esp32spi` and `wiznet5k` and `esp32s3_tft`. I used the code from [here](https://github.com/adafruit/Adafruit_CircuitPython_ConnectionManager/pull/1) under `MQTT example`

brentru:
@justmobilize  Thanks for testing on hardware. Marking this as approved!

brentru (approved):

PR #199, opened 2024-01-24
improve ping handling

vladak:
This changes the way how pings are treated. Rather than sending PINGREQ based on how `loop()` is called, it is based on how messages are sent to the broker (in accordance with the MQTT spec). Also, the ping handling was folded into the loop in `loop()` to make it work in cases with loop timeouts longer than the keep alive timeout.

I plan to test this on a QtPy.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -226,7 +226,7 @@ class MQTT:
         self._is_connected = False
         self._msg_size_lim = MQTT_MSG_SZ_LIM
         self._pid = 0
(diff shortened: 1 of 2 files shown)


FoamyGuy (approved):
This looks good to me. 

I merged main and resolved the conflicts. 

I tried testing with the given test and did get it to run part way but I think the second half of it doesn't work with adafruit.io in the same way as your test broker.  on AIO I hit rate limits that the front end warned me about and the PINGRESP was not sent as the script expected.

I did test basic MQTT functionality seperately and both receiving and sending data appears to be working successfully. 

My testing was on a feather TFT ESP32-S3

PR #203, opened 2024-02-10
Simplify socket exceptions

justmobilize:
The purpose of this PR is to simplify socket exceptions, so the move to [ConnectionManager](https://github.com/adafruit/Adafruit_CircuitPython_ConnectionManager) is smother.

Currently when getting a socket, an `OSError` is not treated as a `TemporaryError` and thus the retry is delayed. As far as I can tell, there is no benefit to delaying the next try here and it will cause `ConnectionManager` to be overly complex.

This treats `OSError` like any other error, and thus no custom handling will be needed in `ConnectionManager`.

If there is something I am missing, please let me know.

### DIFF
--- a/tests/test_port_ssl.py
+++ b/tests/test_port_ssl.py
@@ -20,7 +20,7 @@ class PortSslSetup(TestCase):
     def test_default_port(self) -> None:
         """verify default port value and that TLS is not used"""
         host = "127.0.0.1"
-        port = 1883
+        expected_port = 1883
 
         with patch.object(socket.socket, "connect") as connect_mock:
             ssl_context = ssl.create_default_context()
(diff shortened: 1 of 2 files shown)


tannewt:
Looks fine to me. Will leave it up to Brent though since he uses this code afaik.

tests/test_port_ssl.py
```diff
@@ -20,7 +20,7 @@ class PortSslSetup(TestCase):
     def test_default_port(self) -> None:
         """verify default port value and that TLS is not used"""
         host = "127.0.0.1"
-        port = 1883
+        expected_port = 1883
```
brentru, line 23: @justmobilize  Why is this variable changed?

brentru (changes requested):
Looks good and I understand the changes. One question:

tests/test_port_ssl.py
justmobilize: @brentru it wasn't being passed in to anything and just used below as `expected_port = port` which I than took out.

brentru: Okay, makes sense.

brentru (approved):

PR #205, opened 2024-02-18
Add `timeout` value.

LelandSindt:
Add `timeout` value to avoid `MMQTTException: loop timeout (0) must be bigger than socket timeout (1))` error.

### DIFF
--- a/examples/native_networking/minimqtt_adafruitio_native_networking.py
+++ b/examples/native_networking/minimqtt_adafruitio_native_networking.py
@@ -90,7 +90,7 @@ mqtt_client.connect()
 photocell_val = 0
 while True:
     # Poll the message queue
-    mqtt_client.loop()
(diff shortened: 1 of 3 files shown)


LelandSindt:
After upgrading to Circuit Python 8.x I found that  `mqtt_client.loop()` would throw `MMQTTException: loop timeout (0) must be bigger than socket timeout (1))` I had to set `timeout=1` to avoid the error. 

https://github.com/LelandSindt/analoguEnvoy/blob/main/circutpythonMQTT/code.py#L80

FoamyGuy (approved):
Looks good to me. Thank you @LelandSindt 

I tested native_networking/minimqtt_adafruitio_native_networking.py and confirmed the same error as well as the fix from this PR. 

My testing was on Feather ESP32S3 TFT 9.0.0.beta.2

Issue #211, opened 2024-04-08
Errors with MQTT CircuitPython 9 Publish


matthewjerome:
We have an implementation of MiniMQTT that has been quite stable in CircuitPython 8 but started to display multiple issues when we tested in CircuitPython 9 (9.0.3 and below) on multiple ESP32-S3 boards.

The code base sends some API requests over HTTP GET/POST using the `Adafruit_CircuitPython_Requests` library (with no issues) and then continues on to connect to the MQTT server over SSL to send sensor observations every few seconds. We tried updating to use `adafruit_connection_manager` for `socket_pool` and `ssl_context` but this does not seem to fix the issue. We also tried to run a `reconnect()` on errors as described in #194 but no luck. It is interesting because sometimes the code runs fine for an hour or more but then most often it causes errors within the first few minutes if not immediately.

We tried various bundles both with `*.py` source code, `*.mpy` bundle pre-built and self-built `*.mpy` variations with no luck and similar observations. The bundle dates we have tried out are : `20240402`, `20240307` and `20240224`.

The most common errors seen (but not necessarily together) are :
```
OSError(11,)
OSError(104,)
MemoryError()
MMQTTException('Repeated connect failures',)
```

Less Common Errors (maybe stemming from the common errors above ? ) :
```
OSError(128,)
<WatchDogTimeout>
gaierror(-2, 'Name or service not known')
```

Sample Code :

```python
import wifi
import adafruit_minimqtt.adafruit_minimqtt as MQTT

radio = wifi.radio
pool = adafruit_connection_manager.get_radio_socketpool(radio)
ssl_context = adafruit_connection_manager.get_radio_ssl_context(radio)

mqtt = MQTT.MQTT(
    broker=...,
    port=...,
    is_ssl=True,
    socket_pool=pool,
    ssl_context=ssl_context,
    username=...,
    password=...,
)

try:
	mqtt.publish(topic, message)
except:
	mqtt.reconnect()
```

Are there any additional debugging steps that we can try out or are there other things we should consider when upgrading to CP9? Thank you!

dhalbert:
There is a core CircuitPython issue open that may match what you're seeing: https://github.com/adafruit/circuitpython/issues/9123. Feel free to contribute to that issue.

Are you using `CIRCUITPY_WIFI_SSID` and `CIRCUITPY_WIFI_PASSWORD`, which will enable the web workflow and cause a connection in advance?

[...]

PR #213, opened 2024-05-13
Don't retry when MQTT response is unauthorized

justmobilize:
Update connection code to break out of the retry look when getting either a `Incorrect username/password` or `Unauthorized` response.

Noticed in #211, although doesn't fix the original post

Tested on a `UM FeatherS3`

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -72,12 +72,18 @@ MQTT_DISCONNECT = b"\xe0\0"
 MQTT_PKT_TYPE_MASK = const(0xF0)
 
 
+CONNACK_ERROR_INCORRECT_PROTOCOL = const(0x01)
+CONNACK_ERROR_ID_REJECTED = const(0x02)
+CONNACK_ERROR_SERVER_UNAVAILABLE = const(0x03)
+CONNACK_ERROR_INCORECT_USERNAME_PASSWORD = const(0x04)
+CONNACK_ERROR_UNAUTHORIZED = const(0x05)
+
 CONNACK_ERRORS = {
-    const(0x01): "Connection Refused - Incorrect Protocol Version",
-    const(0x02): "Connection Refused - ID Rejected",
-    const(0x03): "Connection Refused - Server unavailable",
-    const(0x04): "Connection Refused - Incorrect username/password",
-    const(0x05): "Connection Refused - Unauthorized",
+    CONNACK_ERROR_INCORRECT_PROTOCOL: "Connection Refused - Incorrect Protocol Version",
+    CONNACK_ERROR_ID_REJECTED: "Connection Refused - ID Rejected",
+    CONNACK_ERROR_SERVER_UNAVAILABLE: "Connection Refused - Server unavailable",
+    CONNACK_ERROR_INCORECT_USERNAME_PASSWORD: "Connection Refused - Incorrect username/password",
+    CONNACK_ERROR_UNAUTHORIZED: "Connection Refused - Unauthorized",
 }
 
 _default_sock = None  # pylint: disable=invalid-name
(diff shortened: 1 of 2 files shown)


justmobilize:
@brentru willing to take a look?

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -535,7 +551,7 @@ def _connect(
                 rc = self._sock_exact_recv(3)
                 assert rc[0] == 0x02
                 if rc[2] != 0x00:
-                    raise MMQTTException(CONNACK_ERRORS[rc[2]])
+                    raise MMQTTException(CONNACK_ERRORS[rc[2]], code=rc[2])
```
justmobilize, line 554: Looking further, would any error 0x01-0x05 be retry-able?

justmobilize: @dhalbert any reason you can think of that we would retry on any of these:
```py
CONNACK_ERROR_INCORRECT_PROTOCOL: "Connection Refused - Incorrect Protocol Version",
CONNACK_ERROR_ID_REJECTED: "Connection Refused - ID Rejected",
CONNACK_ERROR_SERVER_UNAVAILABLE: "Connection Refused - Server unavailable",
CONNACK_ERROR_INCORECT_USERNAME_PASSWORD: "Connection Refused - Incorrect username/password",
CONNACK_ERROR_UNAUTHORIZED: "Connection Refused - Unauthorized",
```
The only on that I could see is a maybe would be `CONNACK_ERROR_SERVER_UNAVAILABLE`

dhalbert:
I defer to @brentru on this.

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -428,17 +438,24 @@ def connect(
                 self.logger.warning(f"Socket error when connecting: {e}")
                 backoff = False
             except MMQTTException as e:
-                last_exception = e
+                self._close_socket()
```
justmobilize, line 441: After helping a bunch of people over the last week, I realized users were getting `Repeated connect failures`, and the second+ all errored that the `socket was already open`. This fixes that so it's closed on the next attempt

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -72,12 +72,18 @@
 MQTT_PKT_TYPE_MASK = const(0xF0)
 
 
+CONNACK_ERROR_INCORRECT_PROTOCOL = const(0x01)
+CONNACK_ERROR_ID_REJECTED = const(0x02)
+CONNACK_ERROR_SERVER_UNAVAILABLE = const(0x03)
+CONNACK_ERROR_INCORECT_USERNAME_PASSWORD = const(0x04)
+CONNACK_ERROR_UNAUTHORIZED = const(0x05)
```
dhalbert, line 79: Do you want these to be available to the end-user, or are they strictly internal? If you make them start with an underscore, then they will disappear and will not take space in RAM or the .mpy file.

dhalbert (changes requested):
One question, but I would be very interested if the close socket fixes some folks' problems.

adafruit_minimqtt/adafruit_minimqtt.py
justmobilize: I could imaging wanting to catch the error `CONNACK_ERROR_INCORECT_USERNAME_PASSWORD` to show a specific error and not need to parse the Exception...

justmobilize:
@dhalbert So this shouldn't really solve anything other then getting a better error, and making it easier to debug. Unless the error they get is really retriable (which I asked above)

adafruit_minimqtt/adafruit_minimqtt.py
justmobilize: @dhalbert did my above comment not come through?

dhalbert: Weirdly, no. Got it, use case understood.

dhalbert (approved):

PR #217, opened 2024-05-21
Update comments specific to networking modules

justmobilize:
Final PR in a set. Updating comments in code for what networking modules go down what path to help when troubleshooting.

Updates from:
https://github.com/adafruit/Adafruit_CircuitPython_Wiznet5k/pull/163
https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/pull/206


justmobilize:
@dhalbert tiny comment update

dhalbert (approved):
Looks good. I know you wanted to get rid of "backwards compatible", but without changing the fona and esp_atcontrol libraries and testing them (which is an issue), then we should leave it in for now.

PR #221, opened 2024-07-24
refactor will_set function to match the publish function and allow the msg/payload to be encoded bytes, not just str, int or float.  

ehagerty:
As far as I can understand in reading the specification and talking with our mqtt broker host, there is no reason to disallow the last_will message being sent as bytes (encoded json) in the same way that you are supposed to for all other publish operations.  As the last_will is still 'just' a publish operation, albeit a delayed one, this seems logical.  Perhaps more controversially, I have changed the use of 'payload' as a key name in last_will to also match the publish function and use 'msg' for the same content.  I did so because I believe it will help prevent any future confusion.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -269,38 +269,74 @@ class MQTT:
 
     def will_set(
         self,
-        topic: Optional[str] = None,
-        payload: Optional[Union[int, float, str]] = None,
-        qos: int = 0,
+        topic: str,
+        msg: Union[str, int, float, bytes],
         retain: bool = False,
+        qos: int = 0,
     ) -> None:
         """Sets the last will and testament properties. MUST be called before `connect()`.
 
         :param str topic: MQTT Broker topic.
-        :param int|float|str payload: Last will disconnection payload.
-            payloads of type int & float are converted to a string.
+        :param str|int|float|bytes msg: Last will disconnection msg.
+            msgs of type int & float are converted to a string.
+            msgs of type byetes are left unchanged, as it is in the publish function.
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -4,21 +4,21 @@
 
 repos:
   - repo: https://github.com/python/black
-    rev: 24.2.0
+    rev: 24.4.2
     hooks:
       - id: black
   - repo: https://github.com/fsfe/reuse-tool
-    rev: v1.1.2
```
FoamyGuy, line 11: I don't think we want to change all of these versions in the pre-commit config file. 

Ideally we try to keep these in sync across all of the libraries to make maintenance easier. Then we can use a patch to update them all at once.

However, I do know there are issues using the pinned version of pylint in python 3.12 and updating it to 3.x does allow it to work in that environment. 

We're also in the process of switching the libraries over to use Ruff instead of pylint and Black. 

I think it's best to revert these changes to the pre-commit file for now and the ruff change-over for this can be submitted as a separate PR since it's unrelated to main functional change in this PR.

ehagerty: Hi, thank your, and my apologies for being thick but I don't think I intentionally changed this file.  Are you asking me to make a new pull request or am I just not understanding you?  Sorry for my lack of clarity :-)

FoamyGuy: @ehagerty no worries. It seems that those were changed in your branch accidentally. A new PR is not needed, just changing them back to the original values that they were for now. 

I've added a commit to this branch just now to change them back so it is taken care of. 

I've been reading up in the MQTT spec and I am in agreement with your conclusion that the last message can be a normal message just like any other. I'm going to give this a quick test today to ensure no other issues with it.

PR #223, opened 2024-08-26
Use ruff instead of pylint and black

FoamyGuy:
I got setup locally to test another PR on this repo so I've gone ahead and done the ruff conversion to submit here.


.gitattributes
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: 2024 Justin Myers for Adafruit Industries
+#
+# SPDX-License-Identifier: Unlicense
+
+* text eol=lf
```
tannewt, line 5: Same as the other PR. I think this is too broad and should be limited to some file extensions in this repo.

tannewt (changes requested):
One gitattributes comment. Good otherwise. Thank you!

.gitattributes
FoamyGuy: The latest commit changes this to use a more specific list of extensions.

tannewt (approved):
Thank you!

PR #224, opened 2024-10-08
Close the socket when there is an error using the socket

rbedia:
This fixes two related problems which made reconnecting not work after
the connection to the server gets dropped unexpectedly.

The MQTT client continues sending messages and on ESP32 eventually an
OSError is raised. The client must then reconnect. But reconnect was
failing because the Adafruit Connection Manager was returning the same
broken socket for the server since it hadn't yet been closed by
MiniMQTT.

Explicitly calling disconnect() would also not close the socket because
it tries to send the disconnect packet on the broken socket which raised
an OSError preventing the socket from being closed.


tannewt (approved):
Thank you!

PR #225, opened 2024-10-09
Resolve race condition for UNSUBACK

ch4nsuk3:
When attempting to unsubscribe from a topic it was possible to receive the following error:

`Traceback (most recent call last):
  File "code.py", line 79, in <module>
  File "/lib/adafruit_minimqtt/adafruit_minimqtt.py", line 869, in unsubscribe
MMQTTException: ('invalid message received as response to UNSUBSCRIBE: 0x30', None)
`

This is due to the unsubscribe function assuming the next packet received will either be the UNSUBACK or an error code, rather than accounting for possible PUBLISH packets coming from the MQTT broker that may have already been in flight. Looking at some packet captures taken from the broker side its possible to get this error before the broker even receives the UNSUB request.

This fix essentially sets the unsubscribe function to match the existing behavior in the subscribe function, which ignores publish messages while waiting for the SUBACK.

Additionally I added named constant values for the SUBACK and UNSUBACK opcodes, replacing the 'magic number' values for comparing returned opcodes to increase clarity.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -66,7 +66,9 @@ MQTT_PINGREQ = b"\xc0\0"
 MQTT_PINGRESP = const(0xD0)
 MQTT_PUBLISH = const(0x30)
 MQTT_SUB = const(0x82)
+MQTT_SUBACK = const(0x90)
 MQTT_UNSUB = const(0xA2)
+MQTT_UNSUBACK = const(0xB0)
 MQTT_DISCONNECT = b"\xe0\0"
 
 MQTT_PKT_TYPE_MASK = const(0xF0)
@@ -774,7 +776,7 @@ class MQTT:
                         f"No data received from broker for {self._recv_timeout} seconds."
                     )
             else:
-                if op == 0x90:
+                if op == MQTT_SUBACK:
                     remaining_len = self._decode_remaining_length()
                     assert remaining_len > 0
                     rc = self._sock_exact_recv(2)
@@ -852,7 +854,7 @@ class MQTT:
                         f"No data received from broker for {self._recv_timeout} seconds."
                     )
             else:
-                if op == 176:
+                if op == MQTT_UNSUBACK:
                     rc = self._sock_exact_recv(3)
                     assert rc[0] == 0x02
                     # [MQTT-3.32]
@@ -862,10 +864,10 @@ class MQTT:
                             self.on_unsubscribe(self, self.user_data, t, self._pid)
                         self._subscribed_topics.remove(t)
                     return
-
-                raise MMQTTException(
-                    f"invalid message received as response to UNSUBSCRIBE: {hex(op)}"
-                )
+                if op != MQTT_PUBLISH:
+                    raise MMQTTException(
+                        f"invalid message received as response to UNSUBSCRIBE: {hex(op)}"
+                    )
 
     def _recompute_reconnect_backoff(self) -> None:
         """
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -865,6 +865,8 @@ class MQTT:
                         self._subscribed_topics.remove(t)
                     return
                 if op != MQTT_PUBLISH:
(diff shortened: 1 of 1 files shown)


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -862,10 +864,10 @@ def unsubscribe(  # noqa: PLR0912, Too many branches
                             self.on_unsubscribe(self, self.user_data, t, self._pid)
                         self._subscribed_topics.remove(t)
                     return
-
-                raise MMQTTException(
-                    f"invalid message received as response to UNSUBSCRIBE: {hex(op)}"
-                )
+                if op != MQTT_PUBLISH:
+                    raise MMQTTException(
```
vladak, line 868: you might want to add comment similar to the one in `subscribe()`, explain the rationale.

vladak:
Looks good and make sense as The MQTT 3.1.1 spec says in section *3.10.4 Response*:
> If a Server deletes a Subscription:
>  -  It MUST stop adding any new messages for delivery to the Client [MQTT-3.10.4-2].
>  -  It MUST complete the delivery of any QoS 1 or QoS 2 messages which it has started to send to the Client [MQTT-3.10.4-3].
>   - It MAY continue to deliver any existing messages buffered for delivery to the Client.

So the last bullet applies for this case.

ch4nsuk3:
Good call on the comment. Its been added.

vladak:
It would be nice if the `testdata` packet sequences in https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/main/tests/test_unsubscribe.py were augmented by the PUBLISH+UNSUBACK case.

ch4nsuk3:
I should be able to make these changes soon.

vladak:
This should do it:
```patch
diff --git a/tests/test_unsubscribe.py b/tests/test_unsubscribe.py
index f7bbb21..1dfbb85 100644
--- a/tests/test_unsubscribe.py
+++ b/tests/test_unsubscribe.py
@@ -68,6 +68,49 @@ testdata = [
             + [0x6F] * 257
         ),
     ),
+    # UNSUBSCRIBE responded to by PUBLISH followed by UNSUBACK
+    (
+        "foo/bar",
+        bytearray(
+            [
+                0x30,  # PUBLISH
+                0x0C,
+                0x00,
+                0x07,
+                0x66,
+                0x6F,
+                0x6F,
+                0x2F,
+                0x62,
+                0x61,
+                0x72,
+                0x66,
+                0x6F,
+                0x6F,
+                0xB0,  # UNSUBACK
+                0x02,
+                0x00,
+                0x01,
+            ]
+        ),
+        bytearray(
+            [
+                0xA2,  # fixed header
+                0x0B,  # remaining length
+                0x00,
+                0x01,  # message ID
+                0x00,
+                0x07,  # topic length
+                0x66,  # topic
+                0x6F,
+                0x6F,
+                0x2F,
+                0x62,
+                0x61,
+                0x72,
+            ]
+        ),
+    ),
     # use list of topics for more coverage. If the range was (1, 10000), that would be
     # long enough to use 3 bytes for remaining length, however that would make the test
     # run for many minutes even on modern systems, so 1000 is used instead.
@@ -95,7 +138,7 @@ testdata = [
 @pytest.mark.parametrize(
     "topic,to_send,exp_recv",
     testdata,
-    ids=["short_topic", "long_topic", "topic_list_long"],
+    ids=["short_topic", "long_topic", "publish_first", "topic_list_long"],
 )
 def test_unsubscribe(topic, to_send, exp_recv) -> None:
     """
```
It passes the `test_unpublish` test on a CPython system.

mjmare:
I'm still getting this error with v7.11.2.

vladak:
> I'm still getting this error with v7.11.2.
> 

well, the PR has not been merged yet.

mjmare:
Well that explains it then 🥴
Thanks for the PR.

dhalbert (approved):
Thanks for the fix. I will go ahead and merge this. I was waiting for the suggestions to be incorporated and missed the suggestion to go ahead and merge.

vladak:
Great to see the library getting more robust. Follow-up PR #236 with the test submitted.

Issue #227, opened 2024-11-22
Loop default timeout raises exception


FoamyGuy:
The current default for timeout is 0, but this is lower than the socket timeout which raises an exception. First noted on discord here https://discord.com/channels/327254708534116352/537365702651150357/1308831572170117192
```
MMQTTException: ('loop timeout (0) must be bigger than socket timeout (1))', None)
```

PR #228, opened 2024-11-22
timout 1 instead of 0

FoamyGuy:
@ladyada 
Resolves: #227 

Solution discussed with @brentru on discord. I did try changing the cpython examples to use 0, but that still leads to the exception being raised about being shorter than the socket timeout, so I've left them alone in this PR.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -931,7 +931,7 @@ class MQTT:
         if timeout < self._socket_timeout:
             raise MMQTTException(
                 f"loop timeout ({timeout}) must be >= "
-                + f"than socket timeout ({self._socket_timeout}))"
+                + f"socket timeout ({self._socket_timeout}))"
             )
 
         self._connected()
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -921,7 +921,7 @@ class MQTT:
 
         return ret
 
-    def loop(self, timeout: float = 0) -> Optional[list[int]]:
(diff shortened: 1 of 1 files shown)


dhalbert (changes requested):
Suggestion:
The current code is:
```py
        if timeout < self._socket_timeout:
            raise MMQTTException(
                f"loop timeout ({timeout}) must be bigger "
                + f"than socket timeout ({self._socket_timeout}))"
            )
```
The test is actually <, not <=, so a more precise message would be `... must be >=  ...` instead of `... must be bigger ...`.

Now, in terms of actually working properly, should the loop timeout be absolutely greater than the socket timeout, or just greater or equal?

query for @brentru

brentru:
>  should the loop timeout be absolutely greater than the socket timeout, or just greater or equal?

Greater or equal

FoamyGuy:
Thank you @dhalbert and @brentru, the message is updated with the latest commit to `>=` instead of `bigger`

brentru (approved):

adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -930,7 +930,7 @@ def loop(self, timeout: float = 0) -> Optional[list[int]]:
         """
         if timeout < self._socket_timeout:
             raise MMQTTException(
-                f"loop timeout ({timeout}) must be bigger "
+                f"loop timeout ({timeout}) must be >= "
                 + f"than socket timeout ({self._socket_timeout}))"
```
dhalbert, line 934: grammar fix
```suggestion
                + f"socket timeout ({self._socket_timeout}))"
```

dhalbert (changes requested):

dhalbert (approved):

PR #229, opened 2024-12-29
Update minimqtt_adafruitio_native_networking.py

manchicken:
The newer versions of the miniMQTT library seem to break when  you put the whole key for the feed in there. `foo/feed/group.feed` no longer works ( [this breaks it](https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/blob/ece3e396ccb5504558f6b11423a96d06a6dfb5c9/adafruit_io/adafruit_io.py#L43-L54) ), but instead it needs to be just `group.feed`.


dhalbert:
@brentru This I am unsure about. Could you review? I can include it in the release for the `send` fix.

brentru (approved):
Looks good to me

brentru:
@dhalbert  Changes in the PR look good to me, good idea w/ merging to release with the send fix.

PR #231, opened 2025-01-01
handle partial socket send()'s

### LINKED ISSUE #230: MiniMQTT sends malformed messages when messages are too large
I've been working on a small device which sends small JPEG photos via MQTT, and it's been giving me difficulty because the messages just never seemed to queue properly whether I'm using Adafruit IO or another MQTT broker (RabbitMQ's MQTT plugin, in this instance).

I put together this code snippet which fails for me 100% of the time on both Adafruit IO+ as well as with RabbitMQ's MQTT plugin: https://gist.github.com/manchicken/36b6d530eec03871f5a5e24834a2266a

Here are the logs I get from RabbitMQ (I don't know how to get comparable logs from Adafruit IO+): https://gist.github.com/manchicken/5fcddcc26c54b72a9705eebfd8bebc3e

(Apologies for the horizontal scrolling, I copied those verbatim from the RMQ logs)

It seems like there's something that happens once you get past 2.833KB which causes the messages to be malformatted.

Versions:
- Board: Raspberry Pi Pico 2W (2350)
- CircuitPython firmware version: adafruit-circuitpython-raspberry_pi_pico2_w-en_US-9.2.1
- Adafruit libraries version: adafruit-circuitpython-bundle-9.x-mpy-20241224
- RabbitMQ version: 3.10.8-1.1+deb12u1
- rabbitmq_mqtt: 3.10.8

Please let me know if there's any additional information that would be helpful.

dhalbert:
- Fixes #230.

The library was using `socket.send()` directly, but not checking the return value to see if all the bytes were sent. So for larger packets, a partial send would go unnoticed. Added `_send_bytes()`, which calls `send()` until all bytes are sent, and also handles `EAGAIN`. I could have used `socket.sendall()`, but it doesn't handle EAGAIN.

I tested by sending 2900-byte packets to a no-history Adafruit IO test feed. @manchicken, who submitted #230, also tested.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -461,6 +461,21 @@ class MQTT:
             raise MMQTTException(exc_msg) from last_exception
         raise MMQTTException(exc_msg)
 
+    def _send_bytes(
+        self,
+        buffer: Union[bytes, bytearray, memoryview],
+    ):
+        bytes_sent: int = 0
+        bytes_to_send = len(buffer)
+        view = memoryview(buffer)
+        while bytes_sent < bytes_to_send:
+            try:
+                bytes_sent += self._sock.send(view[bytes_sent:])
+            except OSError as exc:
+                if exc.errno == EAGAIN:
+                    continue
+                raise
+
     def _connect(  # noqa: PLR0912, PLR0915, Too many branches, Too many statements
         self,
         clean_session: bool = True,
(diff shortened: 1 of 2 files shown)


manchicken:
The patch you sent me seems to be working well. Thank you so much for your help, you saved my day!

brentru (approved):
Provided that `sendall` doesn't handle `EAGAIN`, I think the implementation you have for `send_bytes` is perfect for addressing this bug.

dhalbert:
`sendall()` doesn't handle `EAGAIN` specially, surprisingly. It just raises it as an `OSError`.

PR #232, opened 2025-01-02
fix EAGAIN reference in _send_bytes()

vladak:
https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/commit/75f384545d7590c0d92594cc450c97238d27874b added code for `_send_bytes()` with incomplete reference to `EAGAIN`. This change fixes that.


dhalbert (approved):
Thanks! I am surprised this was not caught by the pre-commit checks.

PR #236, opened 2025-01-19
add test for the PUBLISH+UNSUBACK case

vladak:
This is a follow-up to PR #225

### DIFF
--- a/tests/test_unsubscribe.py
+++ b/tests/test_unsubscribe.py
@@ -68,6 +68,49 @@ testdata = [
             + [0x6F] * 257
         ),
     ),
+    # UNSUBSCRIBE responded to by PUBLISH followed by UNSUBACK
+    (
+        "foo/bar",
+        bytearray(
+            [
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Thanks! Tests are always a good idea.

dhalbert:
This PR had a test failure when it was merged: https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/actions/runs/12854246695/job/35838388261. I think the timing test maybe should be made wider

vladak:
will take a look later today.

vladak:
This was another case of `test_recv_timeout.py` failing (i.e. not related to this change), this time with:
```
FAILED tests/test_recv_timeout.py::RecvTimeout::test_recv_timeout_vs_keepalive - assert (225.125585554 - 216.917325518) <= (8 + 0.2)
```
so it looks like it would not hurt bumping the adjustment value to 0.3.

PR #237, opened 2025-01-19
no need to check zero byte returned from recv_into()

vladak:
I believe the zero byte check in `_wait_for_msg()` can be removed as the issue was present in 8.x releases while we are now safely into 9.x territory.

Related: #237


dhalbert (approved):
Great, thanks!

PR #238, opened 2025-01-19
increase test_recv_timeout delay

dhalbert:
There's a small interval in `test_recv_timeout.py` of how long to wait for a response. This was first 0.1 seconds, then 0.2, and then that failed too, by a whisker: https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/pull/236#issuecomment-2600877237. This increases the wait to half a second.

### DIFF
--- a/tests/test_recv_timeout.py
+++ b/tests/test_recv_timeout.py
@@ -46,7 +46,7 @@ class RecvTimeout(TestCase):
                     mqtt_client.ping()
 
                 now = time.monotonic()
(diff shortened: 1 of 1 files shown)


brentru (approved):
This PR addresses https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/pull/236#issuecomment-2600961494 and a timeout of 0.5s would likely always pass the CI (buffer of +/0 0.2s from the recommended value)

PR #239, opened 2025-01-20
bump the tolerance for keep alive value in test_recv_timeout.py to 0.3s

vladak:
As discussed in https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/pull/236#issuecomment-2600877237 this raises the tolerance of the keep alive value in the `test_recv_timeout.py` test even further to avoid test failures.


vladak:
closing in favor of #238

PR #245, opened 2025-02-24
Remove secrets usage

justmobilize:
Remove usage of secrets and more:
1. change from using `secrets.py` to `os.getenv`
2. switch from `ESPSPI_WiFiManager` to `WiFiManager`
3. update pool, ssl_context setting to use `adafruit_connection_manager` like all the other examples
4. Updated some things to use f-strings

### DIFF
--- a/examples/cpython/minimqtt_simpletest_cpython.py
+++ b/examples/cpython/minimqtt_simpletest_cpython.py
@@ -8,12 +8,15 @@ from os import getenv
 import adafruit_minimqtt.adafruit_minimqtt as MQTT
 
 # Add your Adafruit IO username and key to your env.
+# (visit io.adafruit.com if you need to create an account, or if you need your Adafruit IO key.)
 # example:
 # export ADAFRUIT_AIO_USERNAME=your-aio-username
 # export ADAFRUIT_AIO_KEY=your-aio-key
+# export broker=io.adafruit.com
 
 aio_username = getenv("ADAFRUIT_AIO_USERNAME")
 aio_key = getenv("ADAFRUIT_AIO_KEY")
+broker = getenv("broker", "io.adafruit.com")
 
 ### Topic Setup ###
 
@@ -23,7 +26,7 @@ mqtt_topic = "test/topic"
 
 # Adafruit IO-style Topic
 # Use this topic if you'd like to connect to io.adafruit.com
-# mqtt_topic = aio_username + "/feeds/temperature"
+# mqtt_topic = f"{aio_username}/feeds/temperature"
 
 
 ### Code ###
@@ -63,7 +66,7 @@ def message(client, topic, message):
 
 # Set up a MiniMQTT Client
 mqtt_client = MQTT.MQTT(
-    broker="io.adafruit.com",
+    broker=broker,
     username=aio_username,
     password=aio_key,
     socket_pool=socket,
--- a/examples/cpython/minimqtt_simpletest_cpython.py
+++ b/examples/cpython/minimqtt_simpletest_cpython.py
@@ -1,17 +1,19 @@
 # SPDX-FileCopyrightText: 2021 ladyada for Adafruit Industries
 # SPDX-License-Identifier: MIT
 
-import os
 import socket
 import ssl
+from os import getenv
 
 import adafruit_minimqtt.adafruit_minimqtt as MQTT
 
-# Add settings.toml to your filesystems. Add your Adafruit IO username and key as well.
-# DO NOT share that file or commit it into Git or other source control.
+# Add your Adafruit IO username and key to your env.
+# example:
+# export ADAFRUIT_AIO_USERNAME=your-aio-username
+# export ADAFRUIT_AIO_KEY=your-aio-key
 
-aio_username = os.getenv("aio_username")
-aio_key = os.getenv("aio_key")
+aio_username = getenv("ADAFRUIT_AIO_USERNAME")
+aio_key = getenv("ADAFRUIT_AIO_KEY")
 
 ### Topic Setup ###
 
@@ -61,7 +63,7 @@ def message(client, topic, message):
 
 # Set up a MiniMQTT Client
 mqtt_client = MQTT.MQTT(
-    broker=os.getenv("broker"),
+    broker="io.adafruit.com",
     username=aio_username,
     password=aio_key,
     socket_pool=socket,
@@ -76,17 +78,17 @@ mqtt_client.on_unsubscribe = unsubscribe
 mqtt_client.on_publish = publish
 mqtt_client.on_message = message
 
-print("Attempting to connect to %s" % mqtt_client.broker)
+print(f"Attempting to connect to {mqtt_client.broker}")
 mqtt_client.connect()
 
-print("Subscribing to %s" % mqtt_topic)
+print(f"Subscribing to {mqtt_topic}")
 mqtt_client.subscribe(mqtt_topic)
 
-print("Publishing to %s" % mqtt_topic)
+print(f"Publishing to {mqtt_topic}")
 mqtt_client.publish(mqtt_topic, "Hello Broker!")
 
-print("Unsubscribing from %s" % mqtt_topic)
+print(f"Unsubscribing from {mqtt_topic}")
 mqtt_client.unsubscribe(mqtt_topic)
 
-print("Disconnecting from %s" % mqtt_client.broker)
+print(f"Disconnecting from {mqtt_client.broker}")
 mqtt_client.disconnect()
--- a/examples/esp32spi/minimqtt_certificate_esp32spi.py
+++ b/examples/esp32spi/minimqtt_certificate_esp32spi.py
@@ -1,8 +1,6 @@
 # SPDX-FileCopyrightText: 2021 ladyada for Adafruit Industries
 # SPDX-License-Identifier: MIT
 
-from os import getenv
-
 import adafruit_connection_manager
 import board
 import busio
--- a/examples/esp32spi/minimqtt_certificate_esp32spi.py
+++ b/examples/esp32spi/minimqtt_certificate_esp32spi.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: 2021 ladyada for Adafruit Industries
 # SPDX-License-Identifier: MIT
 
+from os import getenv
+
 import adafruit_connection_manager
 import board
 import busio
--- a/examples/esp32spi/minimqtt_certificate_esp32spi.py
+++ b/examples/esp32spi/minimqtt_certificate_esp32spi.py
@@ -10,14 +10,14 @@ from digitalio import DigitalInOut
 
 import adafruit_minimqtt.adafruit_minimqtt as MQTT
 
-### WiFi ###
+# Get WiFi details and MQTT keys, ensure these are setup in settings.toml
+ssid = getenv("CIRCUITPY_WIFI_SSID")
+password = getenv("CIRCUITPY_WIFI_PASSWORD")
+broker = getenv("broker")
+username = getenv("username")
+paswword = getenv("paswword")
 
-# Get wifi details and more from a secrets.py file
-try:
-    from secrets import secrets
-except ImportError:
-    print("WiFi secrets are kept in secrets.py, please add them there!")
-    raise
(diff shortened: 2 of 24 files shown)


justmobilize:
There are 6 PRs that use `ESPSPI_WiFiManager` and are the largest user of `secrets` in examples. This is 1 of them

dhalbert (approved):
This looks OK to me but as this is quite a vital library it would be good for @brentru and/or @tyeth to review also.

examples/cpython/minimqtt_simpletest_cpython.py
```diff
@@ -61,7 +63,7 @@ def message(client, topic, message):
 
 # Set up a MiniMQTT Client
 mqtt_client = MQTT.MQTT(
-    broker=os.getenv("broker"),
+    broker="io.adafruit.com",
```
tyeth, line 66: Maybe we stick with broker env var lookup and fall back to io.adafruit.com using the second parameter / default value, as this example isn't really IO related (just a good test server if the user already has an account). Similarly here: https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/pull/245/files#diff-afdace9507c758aee40533c5ef2f1cacbfe152f00f601e2fc07571931dfa63f8R16 which is `examples/esp32spi/minimqtt_certificate_esp32spi.py` line 16.

tyeth (commented):
Looks great. Only a couple of thoughts related to beginners, and using the broker address from os.getenv with a default value of io.adafruit.com wherever `broker` is used.
 I'd quite like to run all of the examples before release, just to be sure, but won't likely get around to it until the end of the week.

examples/esp32spi/minimqtt_simpletest_esp32spi.py
```diff
@@ -11,12 +11,12 @@
 
 import adafruit_minimqtt.adafruit_minimqtt as MQTT
 
-# Add settings.toml to your filesystem CIRCUITPY_WIFI_SSID and CIRCUITPY_WIFI_PASSWORD keys
-# with your WiFi credentials. Add your Adafruit IO username and key as well.
-# DO NOT share that file or commit it into Git or other source control.
```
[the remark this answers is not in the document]
justmobilize: So these are copied from other example. Like Requests just has:
```
# Get WiFi details, ensure these are setup in settings.toml
ssid = os.getenv("CIRCUITPY_WIFI_SSID")
password = os.getenv("CIRCUITPY_WIFI_PASSWORD")
```
I did my best to merge them. Happy to update these to have the warnings if wanted.

examples/cpython/minimqtt_simpletest_cpython.py
justmobilize: @tyeth , so do:
```
broker = getenv("broker", "io.adafruit.com")
```
in this one?

examples/esp32spi/minimqtt_simpletest_esp32spi.py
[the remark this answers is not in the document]
justmobilize: @tyeth, as far as `os.getenv` vs `getenv`, I'm good either way. The examples across the libraries are a mix. Let me know what you want here

examples/cpython/minimqtt_simpletest_cpython.py
tyeth: yeah, something like that. think i spotted "broker" in a few other places too (just glacing at the connect lines)

tyeth:
Likewise, not too fussed, just thought worth the mention, and if memory
servers the guides mostly refer to os.getenv and then link to the examples
from github, but I only really recollect it in the "settings.toml /
internet connect" learn guide template page that most reuse.
Leave it to @brentru or @dhalbert to express a preference (or more likely
just leave as is), and same with the beginner warning about not
committing the settings file.

examples/cpython/minimqtt_simpletest_cpython.py
justmobilize: Will find all of these and make sure they are updated correctly. Will hold until I get other answers to do the work

examples/esp32spi/minimqtt_simpletest_esp32spi.py
```diff
@@ -91,10 +91,9 @@ def message(client, topic, message):
 
 # Set up a MiniMQTT Client
 mqtt_client = MQTT.MQTT(
-    broker=os.getenv("broker"),
-    port=os.getenv("port"),
-    username=os.getenv("username"),
-    password=os.getenv("password"),
+    broker="io.adafruit.com",
```
tyeth, line 94: This example is actually meant to be optionally another server and port (just as it's simple test we might as well show the optional arg, you can use 1883 for insecure or 8883 for secure), as all it does it test callbacks and no IO specific behaviour.

examples/esp32spi/minimqtt_pub_sub_blocking_topic_callbacks_esp32spi.py
```diff
@@ -94,8 +93,8 @@ def on_message(client, topic, message):
 
 # Set up a MiniMQTT Client
 client = MQTT.MQTT(
-    broker=os.getenv("broker"),
-    port=os.getenv("broker_port"),
+    broker=broker,
+    port=broker_port,
     socket_pool=pool,
```
tyeth, line 98: Think we want to add the username and password here too, not sure how this ever worked unless it was a broker that doesn't require one (like test.mosquitto.org), but the aio_username used in topic later is now needed so we might as well offer it here too.
```suggestion
    broker=broker,
    port=broker_port,
    username=aio_username,
    password=aio_key,
    socket_pool=pool,
```

examples/esp32spi/minimqtt_pub_sub_blocking_topic_callbacks_esp32spi.py
```diff
@@ -106,14 +105,14 @@ def on_message(client, topic, message):
 client.on_subscribe = subscribe
 client.on_unsubscribe = unsubscribe
 client.on_message = on_message
-client.add_topic_callback(secrets["aio_username"] + "/feeds/device.batterylevel", on_battery_msg)
+client.add_topic_callback(aio_username + "/feeds/device.batterylevel", on_battery_msg)
```
tyeth, line 108: This doesn't exist anymore

examples/esp32spi/minimqtt_pub_sub_blocking_topic_callbacks_esp32spi.py
```diff
@@ -13,14 +13,13 @@
 
 import adafruit_minimqtt.adafruit_minimqtt as MQTT
 
-### WiFi ###
+# Get WiFi details and broker keys, ensure these are setup in settings.toml
+ssid = getenv("CIRCUITPY_WIFI_SSID")
+password = getenv("CIRCUITPY_WIFI_PASSWORD")
+broker = getenv("broker")
+broker_port = getenv("broker_port")
```
tyeth, line 21: Maybe we provide defaults here too for port/broker, in addition to getting the aio_username (and I've grabbed key too)

```suggestion
ssid = getenv("CIRCUITPY_WIFI_SSID")
password = getenv("CIRCUITPY_WIFI_PASSWORD")
aio_username = getenv("ADAFRUIT_AIO_USERNAME")
aio_key = getenv("ADAFRUIT_AIO_KEY")
broker = getenv("broker", "io.adafruit.com")
broker_port = getenv("broker_port", 1883) # Port 1883 insecure, 8883 secure

```

justmobilize:
@tyeth I updated all the places we needed `broker`, fixed some `f-strings`, an esp32spi update from moving rssi (to match `wifi.radio` a while ago) and removed unsecure ports when we are passing in a ssl context.

examples/esp32spi/minimqtt_simpletest_esp32spi.py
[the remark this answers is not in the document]
justmobilize: @tyeth I'm happy with what ever you want here. Let me know

examples/esp32spi/minimqtt_certificate_esp32spi.py
```diff
@@ -10,14 +10,14 @@
 
 import adafruit_minimqtt.adafruit_minimqtt as MQTT
 
-### WiFi ###
+# Get WiFi details and MQTT keys, ensure these are setup in settings.toml
+ssid = getenv("CIRCUITPY_WIFI_SSID")
+password = getenv("CIRCUITPY_WIFI_PASSWORD")
+broker = getenv("broker")
+username = getenv("username")
+paswword = getenv("paswword")
```
tyeth, line 18: getenv not defined

tyeth (changes requested):
Just that one missing import, apart from that looks great, thanks for all the hard work!

I can't deny I wanted to suggest the "broker" environment variable be uppercase, but seeing the older fona example use it I guess either we have that as the odd one out or just leave broker alone for now.

examples/esp32spi/minimqtt_certificate_esp32spi.py
justmobilize: @tyeth good catch. Fixed.

Ruff didn't catch that, so that's strange (cc @FoamyGuy)

justmobilize:
> Just that one missing import, apart from that looks great, thanks for all the hard work!
> 
> I can't deny I wanted to suggest the "broker" environment variable be uppercase, but seeing the older fona example use it I guess either we have that as the odd one out or just leave broker alone for now.

Happy to update, but the few places in the bundle it exists are all lowercase

tyeth:
Leave it all lowercase.

I'm less impressed that it all passed tests but with broken examples.
This is my first direct experience of ruff Vs pylint, although discussed a
lot with peers [seems the rule categories are insanely grouped and I
dislike using numbers to disable rules as unintelligible] and can only
assume that our rules are not optimally chosen yet.
@FoamyGuy have you seen this behaviour before?

justmobilize:
@tyeth are you good with the code?

I think we can bring up the ruff issues in discord

tyeth:
I haven't seen much traction in discord for some things, probably as there are too many notifications in the circuitpython-dev channel to make it a pleasant experience catching up on all discussions (and my questions being poorly framed or with no definitive answer - plus I should maybe tag people even if occasionally unwanted).

I'm happy to merge this, but I think I want a real issue raised that ruff is not working as expected possibly meaning errors are being introduced without automatic detection.
It would be helpful to have the CI show the changed files as far as pre-commit is concerned, as I can't tell which files it's run ruff against. This makes me think the issue should be raised in https://github.com/adafruit/workflows-circuitpython-libs (?) maybe with this series of commits and CI runs as the trail to follow.
 Unfortunately if we merge this then we won't be able to see updated output from potential fixes in https://github.com/adafruit/workflows-circuitpython-libs although surely similar testing is possible.

Also please forgive my unfamiliarity of the circuitpython workings and processes, I normally live in arduino land and keep half an ear on circuitpython only occasionally do a bit of fixing/testing/maintenance (along with some weekend projects).

tyeth (approved):

justmobilize:
@tyeth I'll dig into the ruff stuff and make sure it comes up next week during `in-the-weeds`

tyeth:
Yeah that's probably best place for it.
I might just poke the rust pre-commit version and re-enable the error to see if it will pick it up...

Plus reorder the ruff format to after fix, to comply with the newest recommendation in 0.9.9 (thanks @Neradoc )
https://github.com/astral-sh/ruff-pre-commit/blob/2c8dce6094fa2b4b668e74f694ca63ceffd38614/README.md?plain=1#L63-L65

tyeth:
Okay, @justmobilize this is now the real ruff error messages (some rules were ignored without preview mode being on, and we still must be missing basic rules if I had to add F821 Undefined Name).

 I've taken out the class to long one, but the rest might be actually this PR, please take a quick look.

justmobilize:
@tyeth I added ignores for the rest, which are valid for now (although could be cleaned up 1 by 1)

tyeth:
> @tyeth I added ignores for the rest, which are valid for now (although could be cleaned up 1 by 1)

Hero level work, thank you so much! 

Making a mental note to go back and remove those after we clear up the default circuitpython library ruff rules

tyeth (approved):

PR #254, opened 2026-02-12
fix: reconnect(): respect QoS and fail-safe

BMDan:
See the two attached issues for more information.

The alternative to the overly-broad `except` would be to burn some RAM on storing a "true" copy separately somewhere in the class.  This felt like a reasonable compromise to avoid that.

Closes #252
Closes #253

See #255 for a version of this PR that's more limited in scope, if you'd rather.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -240,7 +240,7 @@ class MQTT:  # noqa: PLR0904  # too-many-public-methods
         self._lw_msg = None
         self._lw_retain = False
 
-        # List of subscribed topics, used for tracking
+        # List of subscribed topics and the QoS for each, used for tracking
         self._subscribed_topics: List[tuple[str, int]] = []
         self._on_message_filtered = MQTTMatcher()
 
@@ -1206,7 +1206,7 @@ class MQTT:  # noqa: PLR0904  # too-many-public-methods
         """
         if isinstance(qos_level, int):
             if qos_level < 0 or qos_level > 2:
-                raise NotImplementedError("QoS must be between 1 and 2.")
+                raise NotImplementedError("QoS must be between 0 and 2.")
         else:
             raise ValueError("QoS must be an integer.")
 
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -241,7 +241,7 @@ class MQTT:  # noqa: PLR0904  # too-many-public-methods
         self._lw_retain = False
 
         # List of subscribed topics, used for tracking
-        self._subscribed_topics: List[str] = []
+        self._subscribed_topics: List[tuple[str, int]] = []
         self._on_message_filtered = MQTTMatcher()
 
         # Default topic callback methods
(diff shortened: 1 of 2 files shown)


adafruit_minimqtt/adafruit_minimqtt.py
```diff
@@ -958,21 +959,31 @@ def reconnect(self, resub_topics: bool = True) -> int:
...
+            ret = self.connect(session_id=self.session_id)
+            self.logger.debug("Reconnected with broker")
+
+            if resub_topics and subscribed_topics:
+                self.logger.debug("Attempting to resubscribe to previously subscribed topics.")
+                self._subscribed_topics = []
+                while subscribed_topics:
+                    feed = subscribed_topics.pop()
+                    self.subscribe(*feed)
+        except Exception:
```
vladak: I wonder if the broad exception could be reduced to the MQTT exception ?

vladak (changes requested):
Stashing the topics is fine, to me. I am not comfortable with the broad catch, though.

vladak:
Here's a test case, primarily meant for #253 however can serve as a test for #252 as well:
```diff
diff --git a/tests/test_reconnect.py b/tests/test_reconnect.py
index 52b8c76..f5f73fe 100644
--- a/tests/test_reconnect.py
+++ b/tests/test_reconnect.py
@@ -237,3 +237,71 @@ def test_reconnect_not_connected() -> None:
 
     assert user_data.get("disconnect") == False
     assert mqtt_client._connection_manager.close_cnt == 0
+
+
+def test_reconnect_subscribe_failure() -> None:
+    """
+    Test reconnect() will not lose previously subscribed topics on subscribe
+    failure inside reconnect().
+
+    This is a bit finicky as it relies on reconnect() calling subscribe() for each
+    topic separately and in reverse order. Also, it checks the internal
+    _subscribed_topics variable and assumes it stores the topics-to-be-subscribed
+    rather than already subscribed topics.
+    """
+    logging.basicConfig()
+    logger = logging.getLogger(__name__)
+    logger.setLevel(logging.DEBUG)
+
+    host = "localhost"
+    port = 1883
+
+    mqtt_client = MQTT.MQTT(
+        broker=host,
+        port=port,
+        ssl_context=ssl.create_default_context(),
+        connect_retries=1,
+    )
+
+    mocket = Mocket(
+        bytearray([
+            0x20,  # CONNACK
+            0x02,
+            0x00,
+            0x00,
+            0x90,  # SUBACK
+            0x03,
+            0x00,
+            0x01,
+            0x00,
+            0x00,
+            0x20,  # CONNACK
+            0x02,
+            0x00,
+            0x00,
+            0x90,  # SUBACK
+            0x02,
+            0x00,
+            0x02,
+            0x00,
+            0x90,  # SUBACK to make subscribe to bar fail
+            0x02,
+            0x00,
+            0x03,
+            0x80,
+        ])
+    )
+    mqtt_client._connection_manager = FakeConnectionManager(mocket)
+    mqtt_client.connect()
+
+    mqtt_client.logger = logger
+
+    topics = [("bar", 0), ("foo", 0)]
+    logger.info(f"subscribing to {topics}")
+    mqtt_client.subscribe(topics)
+
+    with pytest.raises(MQTT.MMQTTException):
+        logger.info("reconnecting")
+        mqtt_client.reconnect()
+
+    assert set(mqtt_client._subscribed_topics) == set(topics)
```

adafruit_minimqtt/adafruit_minimqtt.py
vladak: Thanks for the detailed information about the thought process, really appreciated.

I think the key question here is whether anything besides `MMQTTException` being ~thrown~raised from the depths of the library code is expected to be recoverable (in general and also w.r.t. the internal MQTT object state). My take on this is that if there is, it should be wrapped in `MMQTTException`, i.e. I do not see the need for the broad exception catch.

PR #258, opened 2026-05-19
Enforce size limit in publish()

### LINKED ISSUE #257: Size limit logic issue
In the IO_MQTT class `self._msg_size_lim` gets set during initialization to a default value of `MQTT_MSG_SZ_LIM = const(10000000)` here:

https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/main/adafruit_minimqtt/adafruit_minimqtt.py#L188

It has public facing properties to get and set its value here:

https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/7ec01da0571483180495f7482b52697e8c48f0db/adafruit_minimqtt/adafruit_minimqtt.py#L270-L282

But there are no other usages of `self._msg_size_lim` in the class. It's value is not actually used to enforce any size limits like the name would suggest. It appears like maybe this was the beginning of an implementation that didn't get completed. 

There is logic inside of publish() that checks if the size is over the MQTT protocol limit of 256mb here:

https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/blob/7ec01da0571483180495f7482b52697e8c48f0db/adafruit_minimqtt/adafruit_minimqtt.py#L702-L703

But there is no logic enforcing the `_msg_size_lim`, only the protocol max. I think an additional check like this should be used to enforce the configurable size limit:
```
if len(msg) > self._msg_size_lim:
    raise ValueError(f"Message size larger than {self._msg_size_lim} bytes.")
```

If logic like this were added, then the AdafruitIO library could make use of it to resolve https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/issues/135 and surface a descriptive error message when an attempt to publish a message that is too large for AIO is made.

FoamyGuy:
Resolves #257

Adds a check and raise against the user configurable `mqtt_msg` property that holds a message size limit.

This change will help the AdafruitIO library have a better error behavior when attempting to publish a message that is too large.

I also think that `mqtt_msg` property name is kind of confusing for the size, but I have left it alone for now since changing would break the API. Though the properties value seemed to have no effect before this change anyhow.

### DIFF
--- a/adafruit_minimqtt/adafruit_minimqtt.py
+++ b/adafruit_minimqtt/adafruit_minimqtt.py
@@ -669,7 +669,7 @@ class MQTT:  # noqa: PLR0904  # too-many-public-methods
                 )
         return rcs
 
-    def publish(  # noqa:  PLR0912, Too many branches
+    def publish(  # noqa:  PLR0912, PLR0915, Too many branches, Too many statements
         self,
         topic: str,
         msg: Union[str, int, float, bytes],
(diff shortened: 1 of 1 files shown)


brentru (approved):
@FoamyGuy  LGTM and addresses the issue at-hand. If you want to go further on the `msg`, I think it's a valid concern, maybe that'd be best addressed in a subsequent PR.

